Issue #3, opened 2022-01-15
New PHP syntax worth considering


KorvinSzanto:
I've gone through the PHP versions since 7.3 which was the most recently released when PSR-12 was accepted and gathered a list of syntax that we may want to cover in this PER. **This list is likely incomplete so please add a comment if there's more I should add.**

### PHP 7.3: https://www.php.net/manual/en/migration73.php
- [x] #5 Heredoc / Nowdoc indention: https://3v4l.org/lTXHQ
- [x] ~~Array destructuring: https://3v4l.org/85FjD~~
- [x] #18 Trailing commas in function / method calls: `foo(1, 2, 3,)`

### PHP 7.4: https://www.php.net/manual/en/migration74.php
- [x] #19 Class property types: `public int $foo;`
- [x] #17 Short functions: `fn(int $foo) => $foo * 2`
- [x] ~~Numeric literal separator: `$int = 1_000_000;`~~
- [x] ~~`__serialize` and `__unserialize` vs using `Serializable`~~
- [x] ~~Nested ternaries require parenthesis in some situations post 7.4, we should consider using `SHOULD NOT` for nested ternaries.~~

### PHP 8.0: https://www.php.net/manual/en/migration80.php
- [x] #42 Named Arguments: `foo(1, c: 2)` https://3v4l.org/tglTA
- [x] #26 Attributes: https://3v4l.org/veQIX
- [x] ~~Constructor property promotion: https://3v4l.org/2hirZ~~
- [ ] #45 Union types: `Foo|Baz` vs `Foo | Baz`
- [x] #21 Match expressions: https://3v4l.org/2s8uK
- [x] ~~Throw as expression: `$a = $b ?: throw Exception('b is falsy');`~~
- [x] #18 Trailing comma in parameter list: https://3v4l.org/RGf19
- [x] ~~`FILTER_VALIDATE_BOOL` over `FILTER_VALIDATE_BOOLEAN` when using `filter_var`~~

### PHP 8.1: https://www.php.net/manual/en/migration81.php
- [x] ~~Array unpacking with string keys: `$array = [...$otherData, 'b' => 1];`~~
- [x] ~~Named arguments used after argument unpacking: `$result = foo(...$args, namedArg: 123)` https://3v4l.org/PN7kg~~
- [x] #7 Enums: https://3v4l.org/8Dljv
- [x] First class callables: `foo(...)` vs `foo( ... )`
[...]

PR #1, opened 2022-01-13
Initial draft

samdark:
- Initial text is taken from PSR-12
- Meta and intro were adjusted to mention PER
- Some typos (commas mainly) were fixed

<strike>Do not merge before #2</strike>

PR #7, opened 2022-04-25
Add section on enumerations.

Crell:
It occurs to me that there is no mention of constant capitalization anywhere else, but IMO constants in enums really should use CamelCase for consistency.  (Frankly they should elsewhere, and I do use them that way elsewhere.)  I don't know if we want to also mention constants more broadly.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1094,7 +1094,7 @@ $instance = new class extends \Foo implements
 
 ## 9. Enumerations
 
-Enumerations (enums) MUST follow the same guidelines as classes, except where otherwise noted here.
+Enumerations (enums) MUST follow the same guidelines as classes, except where otherwise noted below.
 
 Methods in enums MUST follow the same guidelines as methods in classes.  Non-public methods MUST use `private`
 instead of `protected`, as enums do not support inheritance.
@@ -1105,7 +1105,7 @@ space between the colon and the backing type.  This is consistent with the style
 Enum case declarations MUST use CamelCase capitalization.  Enum case declarations MUST be on their own line.
 
 Constants in Enumerations MAY use either CamelCase or UPPER_CASE capitalization.  CamelCase is RECOMMENDED,
-so that it is consistent with cases.
+so that it is consistent with case declarations.
 
 ```php
 enum Suit: string
--- a/spec.md
+++ b/spec.md
@@ -68,6 +68,17 @@ class Foo extends Bar implements FooInterface
         // method body
     }
 }
+
+Enum Beep: int
+{
+    case A = 1;
+    case B = 2;
+
+    public function isOdd(): bool
+    {
+        return $this->value() % 2;
+    }
+}
 ```
 
 ## 2. General
@@ -233,7 +244,7 @@ declare(ticks=1) {
 
 ## 4. Classes, Properties, and Methods
 
-The term "class" refers to all classes, interfaces, and traits.
+The term "class" refers to all classes, interfaces, traits, and enums.
 
 Any closing brace MUST NOT be followed by any comment or statement on the
 same line.
@@ -1083,20 +1094,18 @@ $instance = new class extends \Foo implements
 
 ## 9. Enumerations
 
-Enumerations MUST follow the same guidelines as classes, except where otherwise noted here.
+Enumerations (enums) MUST follow the same guidelines as classes, except where otherwise noted here.
 
-Methods in Enumerations MUST follow the same guidelines as methods in classes.  Non-public methods MUST use `private`
-instead of `protected`.
+Methods in enums MUST follow the same guidelines as methods in classes.  Non-public methods MUST use `private`
+instead of `protected`, as enums do not support inheritance.
 
-In the case of a backed enum, there MUST NOT be a space between the enum name and colon, and there MUST be exactly one
-space between the colon and the backing type.
+When using a backed enum, there MUST NOT be a space between the enum name and colon, and there MUST be exactly one
+space between the colon and the backing type.  This is consistent with the style for return types.
 
-Enumeration cases MUST use CamelCase capitalization.
+Enum case declarations MUST use CamelCase capitalization.  Enum case declarations MUST be on their own line.
 
 Constants in Enumerations MAY use either CamelCase or UPPER_CASE capitalization.  CamelCase is RECOMMENDED,
 so that it is consistent with cases.
-
-No more than one case is permitted per line.
 
 ```php
 enum Suit: string
--- a/spec.md
+++ b/spec.md
@@ -1081,6 +1081,36 @@ $instance = new class extends \Foo implements
 };
 ```
 
+## 9. Enumerations
+
+Enumerations MUST follow the same guidelines as classes, except where otherwise noted here.
+
+Methods in Enumerations MUST follow the same guidelines as methods in classes.  Non-public methods MUST use `private`
+instead of `protected`.
+
+In the case of a backed enum, there MUST NOT be a space between the enum name and colon, and there MUST be exactly one
+space between the colon and the backing type.
+
+Enumeration cases MUST use CamelCase capitalization.
+
+Constants in Enumerations MAY use either CamelCase or UPPER_CASE capitalization.  CamelCase is RECOMMENDED,
+so that it is consistent with cases.
+
+No more than one case is permitted per line.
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -1076,6 +1076,36 @@ $instance = new class extends \Foo implements
 };
 ~~~
 
+## 9. Enumerations
+
+Enumerations MUST follow the same guidelines as classes, except where otherwise noted here.
+
+Methods in Enumerations MUST follow the same guidelines as methods in classes.  Non-public methods MUST use `private`
```
KorvinSzanto, line 1083: I'm thinking we also adjust the term defined in section 4 to include enums since they follow the same rules.

spec.md
```diff
@@ -1076,6 +1076,36 @@ $instance = new class extends \Foo implements
...
+instead of `protected`.
```
KorvinSzanto, line 1084: It'd be good to include a justification here to make it obvious why this is here. Maybe something like "instead of `protected` as Enumeration inheritance is currently not supported."

spec.md
```diff
@@ -1076,6 +1076,36 @@ $instance = new class extends \Foo implements
...
+
+In the case of a backed enum, there MUST NOT be a space between the enum name and colon, and there MUST be exactly one
+space between the colon and the backing type.
+
+Enumeration cases MUST use CamelCase capitalization.
+
+Constants in Enumerations MAY use either CamelCase or UPPER_CASE capitalization.  CamelCase is RECOMMENDED,
+so that it is consistent with cases.
+
+No more than one case is permitted per line.
```
KorvinSzanto, line 1094: The term "case" appears in this text with three different meanings:
1. "In the case of" referring to a hypothetical situation
2. "CamelCase ... consistent with cases." referring to type casing
3. "No more than one case" referring to `case` declarations.

Let's avoid use #1 by using terminology like "When using a backed enum," and switch to "case declaration" for use #3 to make it more plainly obvious what exactly is meant.

spec.md
```diff
@@ -1076,6 +1076,36 @@ $instance = new class extends \Foo implements
...
+
+```php
```
KorvinSzanto, line 1096: This example should have some kind of lead in like:

> The following is an example of a compliant enum:

spec.md
```diff
@@ -1076,6 +1076,36 @@ $instance = new class extends \Foo implements
...
+Methods in Enumerations MUST follow the same guidelines as methods in classes.  Non-public methods MUST use `private`
+instead of `protected`.
+
+In the case of a backed enum, there MUST NOT be a space between the enum name and colon, and there MUST be exactly one
+space between the colon and the backing type.
+
+Enumeration cases MUST use CamelCase capitalization.
+
+Constants in Enumerations MAY use either CamelCase or UPPER_CASE capitalization.  CamelCase is RECOMMENDED,
+so that it is consistent with cases.
```
KorvinSzanto, line 1092: Let's remove this and leave PSR-1 as the authoritative document for constant capitalization for now. In the future I think we want to consider rolling PSR-1 into this PER so that we can more effectively manage the content there.

spec.md
```diff
@@ -1076,6 +1076,36 @@ $instance = new class extends \Foo implements
...
 ~~~
 
+## 9. Enumerations
+
+Enumerations MUST follow the same guidelines as classes, except where otherwise noted here.
+
+Methods in Enumerations MUST follow the same guidelines as methods in classes.  Non-public methods MUST use `private`
+instead of `protected`.
+
+In the case of a backed enum, there MUST NOT be a space between the enum name and colon, and there MUST be exactly one
```
KorvinSzanto, line 1086: I like using "enum" wherever possible over "enumeration" but we should probably do some introduction to the term in the first sentence of this section like "Enumerations (enums for short) MUST follow..."

spec.md
```diff
@@ -1076,6 +1076,36 @@ $instance = new class extends \Foo implements
...
+space between the colon and the backing type.
```
KorvinSzanto, line 1087: It might be good to add a nod to return type declaration requirements since these are the same.

spec.md
Crell: I don't think we should ignore it; case capitalization is the #1 question about enums I see people asking, moreso than anything else.  Yes, really. :smile:

KorvinSzanto: This comment is specific to the constant casing, I agree we should leave the case casing requirement

Crell: I would still favor addressing both here, as they are, by design, closely related.

spec.md
Crell: Probably, but the other examples in the document do not.  We should go through and add consistent language to all of them in a separate task.

spec.md
```diff
@@ -1081,6 +1092,34 @@ $instance = new class extends \Foo implements
 };
 ```
 
+## 9. Enumerations
+
+Enumerations (enums) MUST follow the same guidelines as classes, except where otherwise noted here.
```
samdark, line 1097:
```suggestion
Enumerations (enums) MUST follow the same guidelines as classes, except where otherwise noted below.
```

spec.md
samdark: What's meant by "with cases" here? What's "cases"? i.e. everything CamelCase or everything UPPER_CASE but not both? Maybe it worth stating it more explicitly.

KorvinSzanto: We should change that to "with case declarations": https://github.com/php-fig/per-coding-style/pull/7#discussion_r862201949

spec.md
```diff
@@ -1081,6 +1092,34 @@ $instance = new class extends \Foo implements
...
+Methods in enums MUST follow the same guidelines as methods in classes.  Non-public methods MUST use `private`
+instead of `protected`, as enums do not support inheritance.
+
+When using a backed enum, there MUST NOT be a space between the enum name and colon, and there MUST be exactly one
+space between the colon and the backing type.  This is consistent with the style for return types.
+
+Enum case declarations MUST use CamelCase capitalization.  Enum case declarations MUST be on their own line.
+
+Constants in Enumerations MAY use either CamelCase or UPPER_CASE capitalization.  CamelCase is RECOMMENDED,
+so that it is consistent with case declarations.
```
mabar, line 1108: I believe you meant PascalCase? 
The Upper-CamelCase is not much used anymore, usually when camelCase is used it means the Lower-camelCase

Crell: I see CamelCase used way more than PascalCase.  The vast majority of the programming world has never used Pascal. :smiley:

samdark:
```
camelCase
snake_case
PascalCase
```

I think @mabar is correct here. First-letter uppercase is PascalCase, not camelCase.

spec.md
```diff
@@ -68,6 +68,17 @@ class Foo extends Bar implements FooInterface
         // method body
     }
 }
+
+Enum Beep: int
+{
+    case A = 1;
+    case B = 2;
```
roxblnfk, line 75:
```suggestion
    case Foo = 1;
    case Bar = 2;
```

There are no difference between PascalCase and UPPER_CASE in the cases.
Please use more than one letter in a word.

spec.md
KorvinSzanto: Yes PSR-1 defines the following:
```
- Class names MUST be declared in `StudlyCaps`.
- Method names MUST be declared in `camelCase`.
```
and PSR-12 adds clarification:
```
The term 'StudlyCaps' in PSR-1 MUST be interpreted as PascalCase where the first letter of each word is capitalized including the very first letter.
```

so let's continue to use these terms where "camel case" means `fooBar` and "pascal case" means `FooBar`

KorvinSzanto (changes requested):
See https://github.com/php-fig/per-coding-style/pull/7#discussion_r922874308

spec.md
samdark, line 1108:
```suggestion
Enum case declarations MUST use CamelCase capitalization.  Enum case declarations MUST be on their own line.

Constants in Enumerations MAY use either PascalCase or UPPER_CASE capitalization. PascalCase is RECOMMENDED,
so that it is consistent with case declarations.
```

samdark:
Changed CamelCase → PascalCase and resolved conflicts.

samdark (approved):

KorvinSzanto (approved):

PR #16, opened 2022-05-23
Add method chaining style specification

vjik:
PR is inspired by https://github.com/yiisoft/docs/issues/155

Placing methods in a chain of methods on separate lines is more readable in most cases. In rare cases, this is not the case, so SHOULD.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -649,6 +649,16 @@ $app->get('/hello/{name}', function ($name) use ($app) {
 });
 ```
 
+Method chaining SHOULD be put on separate lines, where each subsequent line is indented once. When doing so, the first
+method MUST be on the next line.
+
+```php
+$someInstance
+    ->create()
+    ->prepare()
+    ->run();
+```
+
 ## 5. Control Structures
 
 The general style rules for control structures are as follows:
--- a/spec.md
+++ b/spec.md
@@ -649,7 +649,7 @@ $app->get('/hello/{name}', function ($name) use ($app) {
 });
 ```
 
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -649,6 +649,16 @@ $app->get('/hello/{name}', function ($name) use ($app) {
...
 ```
 
+Method chaining SHOULD be put on separate lines, where each subsequent line is indented once. When doing so, the first
+method MUST be on the next line.
+
+```php
+$someInstance
+    ->create()
+    ->prepare()
+    ->run();
```
xepozz, line 659: Placing `;` on a separated line is handy as well.

xepozz: It's just a little handy to comment/move/remove lines when you have `;` not at the end of the chain, but I think it's unnecessary to have it as a recommendation :)

andrewprofile: in my opinion it is best to admit both versions :)

samdark (approved):

KorvinSzanto:
I think we should change the "SHOULD" to be "MAY" and I agree that we should allow the parenthesis to be placed on the next line but maybe we can deal with that addition separately.

spec.md
```diff
@@ -649,6 +649,16 @@ $app->get('/hello/{name}', function ($name) use ($app) {
 });
 ```
 
+Method chaining SHOULD be put on separate lines, where each subsequent line is indented once. When doing so, the first
```
samdark, line 652:
```suggestion
Method chaining MAY be put on separate lines, where each subsequent line is indented once. When doing so, the first
```

samdark:
Replaced SHOULD with MAY.

PR #17, opened 2022-06-01
Add a section on short-lambdas

Crell:
I actually quite detest the extra space after the `fn`, and find it hard to read and write.  However, it is enforced by coding style tools now and is consistent with the long-closure form, which also requires the extra space.

The rest is, I think, uncontroversial.  (I hope.)

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1047,6 +1047,36 @@ $foo->bar(
 );
 ```
 
+### 7.1 Short Closures
+
+Short closures, also known as arrow functions, MUST follow the same guidelines
+and principles as long closures above, with the following additions.
+
+The `fn` keyword MUST be preceded and succeeded by a space.
+
+The `=>` symbol MUST be preceded and succeeded by a space.
+
+The semicolon at the end of the expression MUST NOT be preceded by a space.
+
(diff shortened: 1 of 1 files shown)


samdark (approved):

KorvinSzanto:
I'm really not a fan of that long form short closure but I don't really have any suggestions to improve it for now.

theodorejb:
I think this was a mistake. The fact that PHP-CS-Fixer previously enforced spacing after the `fn` keyword using the same setting as for the `function` keyword  [was a bug](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/issues/5916), and has been corrected now with the addition of a separate `closure_fn_spacing` setting. Is there any other PSR-12 tooling that enforces a space after the `fn` keyword by default? PhpStorm specifically formats arrow functions with no space after the keyword. Note that this is how arrow functions [were intended to be written](https://www.reddit.com/r/PHP/comments/fctb18/comment/fjd9u6a/) per the RFC author.

The primary intention for arrow functions is to avoid much of the verbosity of anonymous functions, and requiring a space after the `fn` keyword is a step backwards from this.

PR #19, opened 2022-06-01
Update the abstract/final section to cover all modifier keywords

Crell:
This covers a couple of new additions: `readonly`, typed properties, etc. Because there's so many, it made sense to me to repurpose the abstract/final/static section to cover the full list, in order.

This may overlap now with section 4.4.  If so, I recommend agreeing on the rules (this PR), then we can refactor the spec text to be better chunked if needed.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -579,13 +579,19 @@ public function process(string $algorithm, &...$parts)
 }
 ```
 
-### 4.6 `abstract`, `final`, and `static`
+### 4.6 Modifier keywords
 
-When present, the `abstract` and `final` declarations MUST precede the
-visibility declaration.
+Properties and methods of a class have numerous keyword modifiers that alter how the
+engine and language handles them.  When present, they MUST be in the following order:
 
-When present, the `static` declaration MUST come after the visibility
-declaration.
+* Inheritance modifier: `abstract` or `final`
+* Visibility modifier: `public`, `protected`, or `private`
+* Scope modifier: `static`
+* Mutation modifier: `readonly`
+* Type declaration
+* Name
+
+All keywords MUST be on a single line, and MUST be separated by a single space.
 
 ```php
 <?php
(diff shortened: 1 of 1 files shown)


samdark (approved):

spec.md
```diff
@@ -579,13 +579,19 @@ public function process(string $algorithm, &...$parts)
 }
 ```
 
-### 4.6 `abstract`, `final`, and `static`
+### 4.6 Modifier keywords
```
KorvinSzanto, line 582:
```suggestion
### 4.6 Modifier Keywords
```

spec.md
```diff
@@ -579,13 +579,19 @@ public function process(string $algorithm, &...$parts)
...
+Properties and methods of a class have numerous keyword modifiers that alter how the
+engine and language handles them.  When present, they MUST be in the following order:
 
-When present, the `static` declaration MUST come after the visibility
-declaration.
+* Inheritance modifier: `abstract` or `final`
+* Visibility modifier: `public`, `protected`, or `private`
+* Scope modifier: `static`
+* Mutation modifier: `readonly`
+* Type declaration
```
KorvinSzanto, line 591: We can probably omit type and name since they can't actually go anywhere else and they technically aren't modifier keywords like the rest are

Crell: While technically that may be true, I think it's easier/clearer for readers if we list out all of the various things that can go before a property/method.  That way there's no ambiguity or external knowledge required about what the language allows vs what the spec encourages.

andrewprofile: yes, you are right but I would describe it as a specific case at this point, not so generally.

andrewprofile: @Crell Sorry my mistake, I meant to add readonly and type declaration as examples, but not to say that it is required in general. I hope you understand now.

Crell: They're not required.  Hence the "when present".  If not present (because you're not using them), they do not apply.

PR #21, opened 2022-06-10
Add match section

JesusValera:
# 📚 Description

Add the `match` section next to `switch`.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -710,7 +710,7 @@ if (
 }
 ```
 
-### 5.2 `switch`, `case`
+### 5.2 `switch`, `case`, `match`
 
 A `switch` structure looks like the following. Note the placement of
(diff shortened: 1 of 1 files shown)


samdark (approved):

KorvinSzanto (changes requested):
This should probably be expanded to more specific about more complex match expressions like:
```php
match (true) {
    ...
}
```
See my example from #3: https://3v4l.org/2s8uK

JesusValera:
I added the following code in the PR:
```php
$returnValue = match ($expr) {
    0 => 'First case',
    1, 2, 3 => multipleCases(),
    default => 'Default case',
};
```

So, `$expr` works for anything (`true` also included).

Which changes do you suggest @KorvinSzanto? 😄

KorvinSzanto (approved):

PR #24, opened 2022-06-23
fix: Unify the usage of () after class instances (including anonymous classes)

fruitl00p:
According to chapter 4:

> ## 4. Classes, Properties, and Methods
>
> The term "class" refers to all classes, interfaces, and traits.
>
> Any closing brace MUST NOT be followed by any comment or statement on the same line.
>
> When instantiating a new class, parentheses MUST always be present even when there are no arguments passed to the constructor.

This MR updates the spec to adhere to that rule on anonymous classes too.
(taken from moved from [php-fig/!1283](https://github.com/php-fig/fig-standards/pull/1283))

The original MR on PHP-Fig also mentioned short closures but [!17](https://github.com/php-fig/per-coding-style/pull/17) already fixes this.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1055,24 +1055,27 @@ in the above section.
 ```php
 <?php
 
-$instance = new class {};
-```
+$instance = new class () {};
+~~~
 
 The opening brace MAY be on the same line as the `class` keyword so long as
 the list of `implements` interfaces does not wrap. If the list of interfaces
 wraps, the brace MUST be placed on the line immediately following the last
 interface.
 
+This also includes the use of parentheses due to directly instantiating the 
+declared class.
+
 ```php
 <?php
 
 // Brace on the same line
-$instance = new class extends \Foo implements \HandleableInterface {
+$instance = new class () extends \Foo implements \HandleableInterface {
     // Class content
 };
 
 // Brace on the next line
-$instance = new class extends \Foo implements
+$instance = new class () extends \Foo implements
     \ArrayAccess,
     \Countable,
     \Serializable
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -1055,24 +1055,27 @@ in the above section.
 ```php
 <?php
 
-$instance = new class {};
-```
+$instance = new class () {};
+~~~
```
andrew-demb, line 1059:
```suggestion
```
```

andrew-demb, line 1059: it should be reverted, isn't it?

Crell:
I'm not actually sure about doing () on an anon class with no constructor.  I don't see that style in practice in the wild, and I don't see how it adds much.

fruitl00p:
I fully understand and might even agree with you but the PSR isn't clear ATM and with tools like php-cs-fixer 'interpretting' the psr as they do, it seems like it will place those parenthesis according to their interpretation.

My interpretation follows their reasoning as mentioned in chapter four and without the specific changes mentioned in the closures section to which anon-classes refers. Thus confusion ;)

TimWolla:
see also php-fig/fig-standards#1206 which:

- Requires the parentheses.
- Does not expect a space before the parentheses.

KorvinSzanto:
I agree with @Crell, I'd prefer seeing this go the other way with a statement allowing the parenthesis be omitted when no arguments are passed.

samdark:
I agree with @Crell and @KorvinSzanto on it.

KorvinSzanto:
@fruitl00p Would you be able to adjust this PR to allow omitting parenthesis when no arguments are passed?

fruitl00p:
Very much.. but am on vacay.. feel free to highjack and get this 'sorted' officially

Crell:
New PR filed to go the other way.

PR #27, opened 2022-07-08
Add First Class Closure rules

Crell:
Not much else to say here.  It's pretty basic, and that's a good thing.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -724,6 +724,12 @@ $someInstance
     ->run();
 ```
 
+### 4.8 Function Callable References
+
+A function or method may be referenced in a way that creates a closure out of it, by providing `...` in place of arguments.
+
+If so, the `...` MUST NOT include any whitespace before or after.  That is, the correct format is `foo(...)`.
+
 ## 5. Control Structures
 
 The general style rules for control structures are as follows:


spec.md
```diff
@@ -1047,6 +1047,13 @@ $foo->bar(
 );
 ```
 
+### 7.1 Function callable references
+
+A function may be referenced in a way that creates a closure out of it, by providing `...` in place of arguments.
```
samdark, line 1052: Need an example.

Crell: There literally is one in the very next sentence. :smile:

samdark: Seems I've missed a PHP feature :)

samdark (approved):

KorvinSzanto:
This looks good to me, I can't help but feel like this belongs in the operators section but I don't really know that the `...` here would be an operator.

Crell:
It's not really an operator.  It's really its own "thing".  If anything, it's an oddball function call syntax, so I suppose it could go there?  I put it in the closures section, but thinking about it, putting it in the function invocation section could make more sense.

Crell:
Rebased, and moved to the function/method section.

kenguest:
Just to tighten up the text, perhaps "The correct format is `foo(...)`." would suffice?

Crell:
It feels like it flows better to me with the connecting clause, but if @KorvinSzanto would favor without it I won't take offense. :-)

KorvinSzanto:
I prefer it without but I prefer to do the finer editing once the entire doc is closer to being ready so we can try to get a common voice.

PR #29, opened 2022-07-17
Fix example and names in `spec.md`

WinterSilence:


WinterSilence:
@samdark ok, fix missed end of example yourself

PR #32, opened 2022-07-18
Update link to RFC 2119

KorvinSzanto:
The link we currently have is 404 for some reason, this new URL is the one I get directed to from google.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -4,7 +4,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
 "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
 interpreted as described in [RFC 2119][].
 
-[RFC 2119]: http://tools.ietf.org/html/rfc2119
+[RFC 2119]: http://datatracker.ietf.org/doc/html/rfc2119
 
 ## Overview
 


samdark (approved):

spec.md
```diff
@@ -4,7 +4,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
 "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
 interpreted as described in [RFC 2119][].
 
-[RFC 2119]: http://tools.ietf.org/html/rfc2119
+[RFC 2119]: http://datatracker.ietf.org/doc/html/rfc2119
```
Jean85, line 7:
```suggestion
[RFC 2119]: https://datatracker.ietf.org/doc/html/rfc2119
```

samdark:
image

That's what I have.

spec.md
KorvinSzanto: The reason you're seeing different behavior is because you're going to the `https` endpoint rather than the `http` endpoint. The `https` one does appear to redirect properly, maybe we can poke IETF to fix the non-https url as well?

As far as switching this URL to `https`, I don't really care either way since there isn't a real privacy concern here. I used `http` in this pull to keep parity with PSRs.

diehlmitchell: leaving my note here,
The link I first found the relevant issue does go to the correct page once you add the extension.

Returns a 404: 
http://tools.ietf.org/html/rfc2119

Returns the relevant doc:
http://tools.ietf.org/html/rfc2119.html

PR #37, opened 2022-08-10
Omit () on zero-argument anonymous classes.

Crell:
Replaces #24

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1187,20 +1187,25 @@ the list of `implements` interfaces does not wrap. If the list of interfaces
 wraps, the brace MUST be placed on the line immediately following the last
 interface.
 
+If the anonymous class has no arguments, the `()` after `class` MUST be omitted.
+
 ```php
 <?php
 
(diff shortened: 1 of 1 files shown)


samdark (approved):

spec.md
```diff
@@ -1187,20 +1187,25 @@ the list of `implements` interfaces does not wrap. If the list of interfaces
...
 
 // Brace on the next line
-$instance = new class extends \Foo implements
+// Constructor arguments
+$instance = new class($a) extends \Foo implements
     \ArrayAccess,
     \Countable,
     \Serializable
 {
+    public function __construct(public int $a) {}
```
KorvinSzanto, line 1208: We don't currently allow the opening and closing brackets be placed on the same line
```suggestion
    public function __construct(public int $a)
    {
    }
```

KorvinSzanto:
It'd be good to update the class instantiation requirement to mention this exception as well since it contradicts this a bit

PR #41, opened 2022-09-12
Clarify that classes should follow the same modifier order as methods.

Crell:
This effectively covers readonly classes in PHP 8.2, because the order of `readonly` and `abstract` is already defined.  It also means any future keywords added to either can follow the same ruleset, so we know exactly what is to be expected.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -633,7 +633,7 @@ public function process(string $algorithm, &...$parts)
 
 ### 4.6 Modifier Keywords
 
-Properties and methods of a class have numerous keyword modifiers that alter how the
+Classes, properties, and methods have numerous keyword modifiers that alter how the
 engine and language handles them.  When present, they MUST be in the following order:
 
 * Inheritance modifier: `abstract` or `final`


samdark (approved):

spec.md
```diff
@@ -633,7 +633,7 @@ public function process(string $algorithm, &...$parts)
 
 ### 4.6 Modifier Keywords
 
-Properties and methods of a class have numerous keyword modifiers that alter how the
+Classes, properties, and methods have numerous keyword modifiers that alter how the
```
vjik, line 636:
```suggestion
Classes, properties and methods have numerous keyword modifiers that alter how the
```

spec.md
```diff
@@ -633,7 +633,7 @@ public function process(string $algorithm, &...$parts)
...
 engine and language handles them.  When present, they MUST be in the following order:
```
vjik, line 637:
```suggestion
engine and language handles them. When present, they MUST be in the following order:
```

spec.md
Crell: The Oxford Comma is the correct way to write English, and omitting is is an error.

PR #42, opened 2022-09-12
Describe named arguments.

Crell:
1. Change "arguments" to "parameters" when describing the function definition.  An argument is what is passed to a parameter.
2. Add a note about named args to the "calls" section.

Ignore previous description, I was dumb. :smile:

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -631,10 +631,6 @@ public function process(string $algorithm, &...$parts)
 }
 ```
 
-When invoking a function or method, the same rules apply for the arguments to the function as for the parameters.
-Additionally, if using named arguments, there MUST NOT be a space between the argument name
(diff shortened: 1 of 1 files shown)


samdark (approved):

spec.md
```diff
@@ -513,12 +513,12 @@ function fooBarBaz($arg1, &$arg2, $arg3 = [])
 }
 ```
 
-### 4.5 Method and Function Arguments
+### 4.5 Method and Function Parameters
```
vjik, line 516: Here describe arguments and parameters both. In PHP used caption [Function arguments](https://www.php.net/manual/en/functions.arguments.php). May be in PER save caption "Method and Function Arguments" also?

func0der (approved):

Issue #22, opened 2022-06-10
Recommend using explicit parentheses in expressions that mix binary boolean operators


TimWolla:
While the operator precedence in PHP is clearly defined so that `&&` has higher precedence than `||` and this precedence matches the vast majority of programming languages in use, expressions mixing those two operators without explicit parentheses still require the reader to know the precedence off the top of their head, adding room for error.

The coding style should require (or recommend) explicit parentheses whenever `&&` and `||` are mixed within an expression to make the intent clear:

```
if ($foo && $bar || $baz) { } // no
if (($foo && $bar) || $baz) { } // yes
if ($foo && ($bar || $baz)) { } // yes, but different semantics than the "no".
```

The ["Disjunctive Normal Form Types" RFC](https://wiki.php.net/rfc/dnf_types) that is currently under discussion explicitly requires parentheses for types that mix union and intersection, likely for the same reason: Removing any possible ambiguity.

There's an open pull request for PHP_CodeSniffer that performs and highlights expressions where these operators are mixed without parentheses:

https://github.com/squizlabs/PHP_CodeSniffer/pull/3205

Anecdotally running the Sniff from the PR on our code base exposed a non-trivial number of bugs where `$foo && ($bar || $baz)` was intended, but `$foo && $bar || $baz` was used, showing that this is an issue that happens in the real world.

KorvinSzanto:
I think this is definitely a good idea to do especially if you have complex conditionals but I don't really think it's something we want to require in this doc as it's more of a question of best practices than readability in my opinion. And I want to avoid filling this document with too many recommendations about best practices.

samdark:
I agree with @KorvinSzanto here. It is not really about just formatting.

TimWolla:
Fair enough.

@samdark Would you mind closing this issue as “not planned”? I believe this improves the usefulness of the closed issue list. You should be able to do so using the dropdown right next to the “Reopen” button.

PR #43, opened 2022-09-12
Disallow too complex of a ternary.

Crell:
I used a SHOULD here rather than a MUST NOT because this is more of a "how you should write" than "how you should style" question.


Crell:
I'm actually going to close this to be consistent with #22.  If we bring back one, we should discuss bringing back the other.

PR #44, opened 2022-09-12
Allow abbreviated empty bodies.

Crell:
Having worked with CPP a ton in the last year and a half, I feel very strongly that we need this.  The rules otherwise require wasting 2 extra lines of pointlessness that does nothing to aid readability.

I'm making this a SHOULD rather than MUST for BC reasons, but I would be OK with making it a MUST.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -513,6 +513,29 @@ function fooBarBaz($arg1, &$arg2, $arg3 = [])
 }
 ```
 
+If a function or method contains no statements (such as a no-op implementation or when using
+constructor property promotion), then the body SHOULD be abbreviated as `{}` and placed on the same
+line as the previous symbol, separated by a space.  For example:
+
+```php
+class Point
+{
+    public function __construct(private int $x, private int $y) {}
+    
+    // ...
+}
+```
+
+```php
+class Point
+{
+    public function __construct(
+      public readonly int $x,
+      public readonly int $y,
+    ) {}
+}
+```
+
 ### 4.5 Method and Function Arguments
 
 In the argument list, there MUST NOT be a space before each comma, and there
--- a/spec.md
+++ b/spec.md
@@ -297,6 +297,14 @@ there are no arguments passed to the constructor.
 new Foo();
 ```
 
+If class contains no additional declarations (such as an exception that exists only extend another exception with a new type),
+then the body of the class SHOULD be abbreviated as `{}` and placed on the same line as the previous symbol,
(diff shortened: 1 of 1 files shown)


KorvinSzanto:
This makes sense to me, the one question I'd pose is would it be better to require the method take up at least two lines? Ex:

```php
class Point
{
    public function __construct(private int $x, private int $y)
    {}
    
    // ...
}
```

```php
class Point
{
    public function __construct(
      public readonly int $x,
      public readonly int $y,
    ) {}
}
```

This feels like it'd help distinguish abbreviated methods from class properties.

Crell:
Honestly, I don't think so.  It doesn't offer any visual benefit for me; it just leaves a `{}` handing there like a forgotten appendix.

We should probably do the same for empty classes, like exceptions.  I'll add that.

spec.md
```diff
@@ -297,6 +297,14 @@ there are no arguments passed to the constructor.
 new Foo();
 ```
 
+If class contains no additional declarations (such as an exception that exists only extend another exception with a new type),
+then the body of the class SHOULD be abbreviated as `{}` and placed on the same line as the previous symbol,
+separated by a space.  For example:
+
+```php
+class MyException extends \RuntimeException {}
```
samdark, line 305: That is inconsistent with regular class declaration and since it's usually places in a separate file, I don't think it worth making this case special.

spec.md
```diff
@@ -513,6 +521,29 @@ function fooBarBaz($arg1, &$arg2, $arg3 = [])
...
+}
+```
+
+```php
+class Point
+{
+    public function __construct(
+      public readonly int $x,
+      public readonly int $y,
+    ) {}
```
samdark, line 543: That is inconsistent with regular method declaration, I don't think it worth making this case special.

samdark (changes requested):

samdark:
While it looks a bit better, it is inconsistent i.e. adds more rules to follow making everything more complicated.

samdark:
Alright. With `MAY` it's OK.

samdark (approved):

Crell:
I'd also be OK with a SHOULD here... :smile:

KorvinSzanto:
I'm still not a fan of method declarations taking a single line, I'd be more happy requiring the `{}` be on the next line to ensure methods don't look too much like class properties.

That said in the interest of progress let's merge this as is and discuss if any one else would like to.

PR #45, opened 2022-09-12
Add handling for compound types.

Crell:
There's likely to be discussion on this, but this PR gets us started.  Discuss.


samdark (approved):

spec.md
```diff
@@ -128,6 +128,55 @@ Any new types and keywords added to future PHP versions MUST be in lower case.
...
+
+function complex(array|(ArrayAccess&Traversable) $input): ArrayAccess&Traversable
+{
+    // ...
+}
+
+function veryComplex(
+    array
+    |(ArrayAccess&Traversable)
+    |(Traversable&Countable) $input): ArrayAccess&Traversable
```
vjik, line 170:
```suggestion
    |(Traversable&Countable) $input
): ArrayAccess&Traversable
```

spec.md
```diff
@@ -128,6 +128,55 @@ Any new types and keywords added to future PHP versions MUST be in lower case.
...
+The following are correct ways to format compound types:
+
+```php
+function foo(int|string $a): User|Product
+{
+    // ...
+}
+
+function somethingWithReflection(
+    \RelfectionObject
```
TimWolla, line 153:
```suggestion
    \ReflectionObject
```

spec.md
Crell: Hm.  Not sure here.  Does this context also force multilining the return part?  If so, that should be said explicitly somewhere.

KorvinSzanto: Yeah doing this would split the argument list over multiple lines and would make "argument list is split across multiple lines" rules apply here. I get that this is different than actually splitting each argument onto their own lines but I don't think that distinction matters at all when it comes to readability and consistency.

spec.md
```diff
@@ -128,6 +128,55 @@ Any new types and keywords added to future PHP versions MUST be in lower case.
 Short form of type keywords MUST be used i.e. `bool` instead of `boolean`,
 `int` instead of `integer` etc.
 
+Compound types includes intersection, union, and mixed intersection and union type declarations. PHP requires
+that all compound types be structured as an ORed (unioned) series of ANDs (intersections), and that each set of
+intersections be encased with parentheses.
+
+The union symbol `|` and intersection symbol `&` MUST NOT have a leading or trailing space.  The parentheses MUST NOT
```
TimWolla, line 135: While I personally agree that there should be no spaces around the "symbols", this appears to be inconsistent with `catch()` which has spaces around in the example:

https://github.com/php-fig/per-coding-style/blob/master/spec.md?plain=1#L967

samdark: That should unified one way or another 👍

KorvinSzanto (changes requested):
Needs:
- Consistency with `catch()` or an argument for not being consistent
- Consistency with mutliline declaration requirements

KorvinSzanto:
I think the catch consistency question is a big one so I've opened a vote in discord for the WG to decide the way to move forward.

Crell:
A while back, I put out a Mastodon poll on this.  Not exactly scientific, but probably the best we've got.

https://phpc.social/@Crell/112496711448592596

That poll at least comes out 60/40 in favor of no-space.  (Some interesting comments either way in the replies.)

Symfony, doesn't make an explicit stance, but their examples also show no-space (link courtesy Ben Ramsey):

https://symfony.com/doc/current/contributing/code/standards.html

And the default configuration for php-cs-fixer also defaults to no-space:

https://cs.symfony.com/doc/rules/whitespace/types_spaces.html#rule-sets

Laravel doesn't say explicitly either, but its examples have no-space in compound types in the docblock:

https://laravel.com/docs/11.x/contributions#coding-style

Based on that, I am strongly leaning toward requiring no-space for compound types, as the PR currently does.  And then probably changing union-catch to match.

WG members (only): Yay :+1: or nay :-1:, based on the above?

Jean85:
> Symfony, doesn't make an explicit stance, but their examples also show no-space (link courtesy Ben Ramsey):
> 
> https://symfony.com/doc/current/contributing/code/standards.html
> 
> And the default configuration for php-cs-fixer also defaults to no-space:
> 
> https://cs.symfony.com/doc/rules/whitespace/types_spaces.html#rule-sets

I should note that the latter is surely causing the former, since Symfony uses PHP-CS-Fixer for its code (it even has a dedicated preset).

samdark:
No-space is used in Yii as well.

Crell:
Based on the above research and lack of objection to it, we will proceed with this PR (which I just rebased) unless I hear an articulated objection in the next week or so.

@KorvinSzanto Please unblock the PR. :smile:

kenguest:
Overall consistency is important, so I would expect the later revision of PER-CS to disallow spaces around "|" for catch statements.

spec.md
```diff
@@ -1003,7 +1052,7 @@ try {
     // ...
 } catch (FirstThrowableType $e) {
     // ...
-} catch (OtherThrowableType | AnotherThrowableType $e) {
+} catch (OtherThrowableType|AnotherThrowableType $e) {
```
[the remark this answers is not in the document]
TimWolla: > for bitwise or usage

Terminology remark: That definitely is _not_ a bitwise or. It can be considered a union type, but also isn't 100% that (because it's not *really* a type declaration in the classic sense).

KorvinSzanto (approved):
I have to approve to remove my request for changes, but I believe this needs to be a 3.0 change given that we're modifying catch

spec.md
```diff
@@ -128,6 +128,55 @@ Any new types and keywords added to future PHP versions MUST be in lower case.
...
+function veryComplex(
+    array
+    |(ArrayAccess&Traversable)
+    |(Traversable&Countable) $input): ArrayAccess&Traversable
+{
+    // ...
+}
+```
+
+If one of the ORed conditions is `null`, it MUST be the last item in the list.
```
vudaltsov, line 176: Why null must be the last? It's kinda inconsistent with `?X`, since there null goes first.

PR #47, opened 2022-09-16
Add short array syntax

mbniebergall:
Added declaring arrays with short array syntax in Keywords and Types section with the "Short form of type keywords MUST ...".


samdark:
#48 seems to be better alternative.

KorvinSzanto:
Closing in favor of #48

PR #48, opened 2022-09-16
Added Arrays as Section 12

mbniebergall:
Added section 12 specifying array use. Section 11 is used by https://github.com/php-fig/per-coding-style/pull/26. Applies guidelines like 2.6 Trailing Commas and 4.7 Method and Function Calls to array.

If this PR is preferred over https://github.com/php-fig/per-coding-style/pull/47 then that one can be declined.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1317,7 +1317,7 @@ function allowed()
 }
 ```
 
-## 12. Arrays
+## 11. Arrays
 
 Arrays MUST be declared using the short array syntax.
 
--- a/spec.md
+++ b/spec.md
@@ -1317,6 +1317,45 @@ function allowed()
 }
 ```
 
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -1317,6 +1317,45 @@ function allowed()
 }
 ```
 
+## 12. Arrays
```
vjik, line 1320:
```suggestion
## 11. Arrays
```

mbniebergall: Updated to 11

samdark (approved):

KorvinSzanto:
I ended up using section 11 for this since that's what was already in use in the PR and since `Ar` comes before `At`, but I suppose that's a bit arbitrary.

PR #50, opened 2022-09-29
Update on how to handle multi lines in control structures

mathroc:
let me know if you think something like this should be done.

the first thing I wanted to take care of was to allow to break a function call or an array into multiple line in a control structures expression without having to switch to full multi line mode for the expression

and I also noticed a case where multi line could be authorized for catch expression with multiple union'ed types


samdark:
I think we should close this PR. Thanks for participating anyway, @mathroc

PR #53, opened 2022-11-10
Update Short Closures spec to remove space after `fn` keyword

theodorejb:
Per the accepted [Arrow Functions RFC](https://wiki.php.net/rfc/arrow_functions_v2) and [per its author](https://www.reddit.com/r/PHP/comments/fctb18/comment/fjd9u6a/), short closures were intended to be written without a space after the `fn` keyword. This is also the default formatting enforced by PhpStorm.

PHP-CS-Fixer was an outlier which incorrectly enforced spacing after the `fn` keyword using the same setting as for the `function` keyword. This was a bug and has been corrected with the addition of a separate `closure_fn_spacing` setting (though this setting still defaults to `one` for backwards compatibility - the plan was to [change it to enforce no space](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/b577444c38b61f2fdc32cd0a4685bd595be391b7/src/Fixer/FunctionNotation/FunctionDeclarationFixer.php#L235) in the next major 4.0 version). But the current PER coding style spec may get in the way of this plan.

Since the primary goal of arrow functions is to avoid the verbosity of anonymous functions, requiring a space after the `fn` keyword is a step backwards, and also doesn't align with general usage promoted by PhpStorm as well as the accepted PHP RFC and documentation.

The section on Short Closures was not added until the 17th of July by https://github.com/php-fig/per-coding-style/pull/17, which was after the 1.0.0 spec was approved/released on June 9, so I'm hoping it's not too late to correct it.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1148,7 +1148,7 @@ $foo->bar(
 Short closures, also known as arrow functions, MUST follow the same guidelines
 and principles as long closures above, with the following additions.
 
-The `fn` keyword MUST be preceded and succeeded by a space.
+The `fn` keyword MUST be preceded by a space, and MUST NOT be succeeded by a space.
 
 The `=>` symbol MUST be preceded and succeeded by a space.
 
@@ -1161,12 +1161,12 @@ The following examples show proper common usage of short closures.
 
 ```php
 
-$func = fn (int $x, int $y): int => $x + $y;
+$func = fn(int $x, int $y): int => $x + $y;
 
-$func = fn (int $x, int $y): int
+$func = fn(int $x, int $y): int
     => $x + $y;
 
-$func = fn (
+$func = fn(
     int $x,
     int $y,
 ): int


Crell:
I quite hate the extra space, so if we can change the rule to not have it (the currently published version doesn't mention it at all, so nothing is finalized yet) I am fully on board.  Hell, I'd get rid of the space on long-closures too, if we could. 😄

jrfnl:
Independently of personal tastes, this introduces an inconsistency in the standard with long closures. As this is supposed to be a coding style standard, consistency is key, so I'm not sure this is a good idea....

theodorejb:
@jrfnl Arrow functions have a unique syntax from full closures, aimed at conciseness, so I don't think it's inconsistent for them to have their own spacing rule.

theodorejb:
Is there anything else I can do to move this forward?

theodorejb:
Note: I also checked the [PHP Prettier plugin](https://github.com/prettier/plugin-php) and found that it, like PhpStorm, formats arrow functions without a space after the `fn` keyword. [Example](https://loilo.github.io/prettier-php-playground/#N4IgDgTgpgLjCWUIHkwIPYDsDOIBcok8mMA6vACYwAW+AHAAwA0IMAhgEblW14AsLAK7YoAFU648AMzYAbES2zEA5rKgBFQehhR8M+VBZhqYAGpIlWfCACsAOj4gWMCG3iyVAYXQBbH2wAFAAkA-BdBQxAOVwBjKABlGABPNWswbAgAWgAmJxBoAEdBeGgA12V-PTkFEGIRCBgytgq2KoMAXxYoCngYdBQ0eCxJUDZsGDaRTtrMMEEJvBAAHgB+YzAAHUwANzYIAH0oAA8wfpgACj3XJP3-MHOpTHOAEl6oH2wASgBeAD4AAiubBu2EEPhebw+nyY-wA2rCAIxMbJMADMTD4AF0mLCbEwAGxMADsTDo2NhAE4mAjmAiEZjMZ9PgBuLYgdpAA).

KorvinSzanto (approved):

KorvinSzanto:
Looks good to me, one thing that pops out at me is the following is incorrect:
> The `fn` keyword MUST be preceded by a space

for example consider something like https://3v4l.org/SRSVo:
```php
foo(fn(string $baz) => "Input: $baz");
// vs
foo( fn(string $baz) => "Input: $baz");
```

but this can be sorted out in a separate PR

jrmajor:
Can we actually make some research on how arrow functions are used in the wild? PSR-12 team conducted [a survey](https://www.php-fig.org/psr/psr-12/meta/#44-public-survey), gathering votes from 142 people including 17 project representatives. For this PR, I can see 4 people, including one being against this change.

> requiring a space after the `fn` keyword (...) doesn't align with general usage

Sure, except for completely niche projects like [Symfony](https://github.com/symfony/symfony/blob/c79d4ab7e787eb27d83581ccb8e15a4e11e88df3/src/Symfony/Component/HtmlSanitizer/HtmlSanitizerConfig.php#L404), [Laravel](https://github.com/laravel/laravel/blob/88b2d177fcff54cc9356d29ba7d3b36efda800c6/database/factories/UserFactory.php#L36), [PHPUnit](https://github.com/sebastianbergmann/phpunit/commit/72e5b3507327ca70ec5e20f6c073e421630ad571), [Composer](https://github.com/composer/packagist/blob/6722b861b02a5bcae7cdc0942a245d4c250fc5f7/src/Entity/PhpStatRepository.php#L88), [Magento](https://github.com/magento/magento2/blob/6dbb7fc48bd05eb2d00c9951b014aea9da7f73d7/app/code/Magento/Customer/Test/Fixture/Customer.php#L191), [Doctrine](https://github.com/doctrine/dbal/blob/82483595da680180c1d360ea7fbff1ea8fb08142/tests/Functional/ResultTest.php#L48), [PHPStan](https://github.com/phpstan/phpstan-src/blob/4cf098d95264610eac35ef64e148fa9486ecd686/src/Type/ErrorType.php#L18), [Flysystem](https://github.com/thephpleague/flysystem/blob/a7790f3dd1b27af81d380e6b2afa77c16ab7e181/src/UrlGeneration/ShardedPrefixPublicUrlGenerator.php#L30), [Laminas](https://github.com/laminas/laminas-code/blob/224f5d71719668c0753a50df55ef6bfb7aa6b587/src/Reflection/FunctionReflection.php#L193), [Phalcon](https://github.com/phalcon/cphalcon/blob/4eb3fcacf4973b62f8fe3bf3baf0dc5a68552a48/tests/unit/Filter/Filter/SanitizeMultipleCest.php#L189), [PHP-CS-Fixer](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/ad0a87e269698a08b0bee4ec5c739790ff23bcff/src/Tokenizer/Tokens.php#L458), [Rector](https://github.com/rectorphp/rector-src/blob/f862bf997aa6cb52d0c2fa3ed75744e4da1b3475/src/Util/ArrayParametersMerger.php#L21), [Spatie](https://github.com/spatie/laravel-medialibrary/blob/3047dfc767fb7b0d67c8d6d4ab2a1cd6885f500d/src/Support/MediaStream.php#L55), [phpDocumentor](https://github.com/phpDocumentor/phpDocumentor/blob/4d5553c2e902546a1b3e2211b46454ef6f7376a6/incubator/json-path/src/Parser/ParserBuilder.php#L50)...

> requiring a space after the `fn` keyword (...) doesn't align with documentation

You mean PHP documentation? It doesn't seem to care about any coding style in the slightest (screenshot from https://www.php.net/manual/en/functions.anonymous.php).

screenshot-2022-11-27T22 03 45@2x

> PHP-CS-Fixer was an outlier which incorrectly enforced spacing after the `fn` keyword

> This is also the default formatting enforced by PhpStorm.

> I also checked the PHP Prettier plugin and found that it, like PhpStorm, formats arrow functions without a space after the fn keyword.

So defaults in PHP-CS-Fixer, which doesn't match your taste, are just incorrect, but bringing up defaults in PhpStorm and Prettier is a valid point in discussion?

> Per the accepted Arrow Functions RFC and per its author, short closures were intended to be written without a space after the fn keyword.

That's actually the only argument that makes sense to me. However, the RFC author did allow a space after the `fn`, I guess in order to let the users decide how they would like to use it, and it seems to me that many of them chose to go with the additional space.

theodorejb:
> Sure, except for completely niche projects like Symfony, Laravel, PHPUnit, Composer, Magento, Doctrine, PHPStan, Flysystem, Laminas, Phalcon, PHP-CS-Fixer, Rector, Spatie, phpDocumentor...

Out of that list, [Symfony](https://github.com/symfony/symfony/blob/c79d4ab7e787eb27d83581ccb8e15a4e11e88df3/.php-cs-fixer.dist.php), [PHPUnit](https://github.com/sebastianbergmann/phpunit/blob/35136a4fb2045460949428bd138c9c091441e5be/.php-cs-fixer.dist.php), [Composer Packagist](https://github.com/composer/packagist/blob/6722b861b02a5bcae7cdc0942a245d4c250fc5f7/.php-cs-fixer.php), [Magento](https://github.com/magento/magento2/blob/6dbb7fc48bd05eb2d00c9951b014aea9da7f73d7/.php-cs-fixer.dist.php), [Flysystem](https://github.com/thephpleague/flysystem/blob/a7790f3dd1b27af81d380e6b2afa77c16ab7e181/.php-cs-fixer.dist.php), [Phalcon](https://github.com/phalcon/cphalcon/blob/4eb3fcacf4973b62f8fe3bf3baf0dc5a68552a48/.php_cs.dist), [PHP-CS-Fixer](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/ad0a87e269698a08b0bee4ec5c739790ff23bcff/.php-cs-fixer.dist.php), and [Spatie](https://github.com/spatie/laravel-medialibrary/blob/3047dfc767fb7b0d67c8d6d4ab2a1cd6885f500d/.php_cs.dist.php) are all using PHP-CS-Fixer, which until the latest release (v3.13) didn't allow differentiating between `function` and `fn` for formatting. So it seems unlikely that these projects chose to use an extra space; they simply didn't have the option to choose something different for `fn` with PHP-CS-Fixer.

Another project in your list ([Laminas](https://github.com/laminas/laminas-code/blob/224f5d71719668c0753a50df55ef6bfb7aa6b587/src/Generator/TypeGenerator.php#L65)) does not appear to enforce an arrow function spacing rule - some `fn` keywords are succeeded by a space, but the majority are not.

Other well-known projects such as [Psalm](https://github.com/vimeo/psalm/blob/9d597cf370b365816eee0207d7f07aac9f203128/src/Psalm/Internal/Analyzer/ScopeAnalyzer.php#L126) enforce having no space after the `fn` keyword.

> You mean PHP documentation? It doesn't seem to care about any coding style in the slightest (screenshot from https://www.php.net/manual/en/functions.anonymous.php).

That screenshot isn't related to arrow functions (I did submit a PR to fix that inconsistency, though). The official [Arrow Functions documentation](https://www.php.net/manual/en/functions.arrow.php) examples do use consistent formatting which matches the accepted RFC.

> So defaults in PHP-CS-Fixer, which doesn't match your taste, are just incorrect, but bringing up defaults in PhpStorm and Prettier is a valid point in discussion?

My point was that PHP-CS-Fixer's defaults didn't match other popular tools. Now that PHP-CS-Fixer allows configuring spacing for `fn` separately, you can already find projects that use it starting to [opt into enforcing no space](https://github.com/swow/swow/blob/abeb82a23e2e693ca11abe5c6b6b387ccc62b082/lib/swow-php-cs-fixer-config/src/Config.php#L106) after `fn`.

Since there hasn't yet been a released PSR standard for `fn` spacing, it's natural that various projects have been using different styles. I have tried to avoid making an argument for one style or another based on my personal taste.

To summarize, the following arguments could be used to favor a space after `fn`:

1. A bunch of large open-source projects use this formatting, since PHP-CS-Fixer defaulted to it and didn't use to allow formatting arrow functions separately from full closures.
2. Some may view it as more consistent with the existing PSR standard for closure spacing.

On the other hand, the following points favor no space after `fn`:

1. The primary goal of arrow functions is to avoid the verbosity of full closures. Not having a space after `fn` supports this goal.
2. Arrow functions were intended to be written without a space after `fn`. This is the formatting consistently used in the accepted RFC, as well as the official PHP documentation for arrow functions.
3. Other popular PHP formatters such the one built into PhpStorm, and Prettier's PHP plugin default to no space after `fn`. I don't have access to direct numbers, but my hunch is that PhpStorm is used with far more projects than PHP-CS-Fixer is.

jrmajor:
> A bunch of large open-source projects use this formatting, since PHP-CS-Fixer defaulted to it and didn't use to allow formatting arrow functions separately from full closures.

> it seems unlikely that these projects chose to use an extra space; they simply didn't have the option to choose something different for `fn` with PHP-CS-Fixer.

They did have the option, it's as simple as `'function_declaration' => false`. They aren't forced to use neither this rule nor the PHP-CS-Fixer itself, yet they've chosen to do so.

> Now that PHP-CS-Fixer allows configuring spacing for `fn` separately, you can already find projects that use it starting to opt into enforcing no space after `fn`.

This option was released almost a month ago, and none of the big projects I brought up changed it so far.

Why do you think that all of these projects have been forced to use this spacing by PHP-CS-Fixer? What makes you sure that any of these projects wants to change that? All of them were free to format their code however they want. They've chosen. They can change it whenever they want, yet they don't.

> You mean PHP documentation? It doesn't seem to care about any coding style in the slightest
> > That screenshot isn't related to arrow functions

It wasn't supposed to be related to arrow functions, it was supposed to show my point: PHP documentation doesn't care about any coding style, so bringing it up in a discussion about coding style makes no sense.

> My point was that PHP-CS-Fixer's defaults didn't match other popular tools.

> Other popular PHP formatters such the one built into PhpStorm, and Prettier's PHP plugin default to no space after `fn`. I don't have access to direct numbers, but my hunch is that PhpStorm is used with far more projects than PHP-CS-Fixer is.

PhpStorm may be used in far more projects, but not as a formatter. PhpStorm can't even be used to enforce consistent formatting, as you can't use it in your CI. And for Prettier we have numbers: 1.5k stars for Prettier, 11.5k stars for PHP-CS-Fixer. So my question remains: why do you call defaults in PHP-CS-Fixer “incorrect”, but bring up defaults in PhpStorm and Prettier is a valid point in discussion?

> The primary goal of arrow functions is to avoid the verbosity of full closures. Not having a space after `fn` supports this goal.

verbose: containing more words than necessary (Merriam-Webster)
verbose: using or containing more words than are necessary (Cambridge)

Sorry for taking apart your words, but I think there's a point there: having more whitespace is something different than having more words, and it makes code easier to read for some people. You don't need to add any spaces around binary operators, after the type colon, or between function arguments, yet you do for some reason. You could probably even format most of your PHP files to be one line to reduce “verbosity”.

> Arrow functions were intended to be written without a space after `fn`. This is the formatting consistently used in the accepted RFC, as well as the official PHP documentation for arrow functions.

I agree with the point about the RFC. The documentation argument doesn't make any sense though, for the reason stated above. There is anything but coding style consistency in PHP documentation.

> Another project in your list (Laminas) does not appear to enforce an arrow function spacing rule
> Other well-known projects such as Psalm enforce having no space after the `fn` keyword.

You're right, my bad with Laminas. Also, I'm not saying that nobody prefers this style, Nette seems to use it too.

My point with bringing up these projects is that **it's just false that `fn(` is a “general usage”**. Many major projects do use `fn (`, and code search on GitHub reveals that it's even more popular than `fn(`.

- **Two most popular PHP frameworks (Symfony, Laravel) use `fn (`**
- **Two most popular testing frameworks (PHPUnit, Pest) use `fn (`**
- **The most popular static analysis tool (PHPStan) uses `fn (`**
- **The most popular package manager (Composer) uses `fn (`**
- **There are more occurrences of `fn (` than `fn(` found with a simple GitHub search**

Crell:
> This option was released almost a month ago, and none of the big projects I brought up changed it so far.

You vastly under-estimate the effort involved in changing coding standards on a large project.  IME, it can take months just to agree to do it, to say nothing of actually making the change.

@jrmajor is correct that we should be informed by research, however.  Which he has, graciously, done for us.  If we take the screenshot from GitHub at face value, it tells us that while `fn (` is more common, it's not by much.  It's around 53% `fn (`, 47% `fn(`.  I'd call that close enough to an even split.  That means no matter what we do, "around half" of existing code will not already be in compliance with PER-CS 1.1.

It's also a valid observation that it seems very likely that many/most of those projects didn't make an active, deliberate decision.  They just turned on php-cs-fixer, or Prettier, or whatever tool and left it at its defaults, or made only one or two changes.  That's true for both styles, probably.

So if we make a statement at all, it's going to disrupt about the same amount of code either way.  And odds are, those people will upgrade their formatting tool at some point in the future, it will say "now using PER-CS 1.1", and flag a bunch of things.  Users will roll their eyes, push "yes", and move on with their lives.

So our decision should be based on internal consistency within the spec, and on what is most ergonomic (which is admittedly hard to get any objective data about in this case).  The popularity argument is basically a wash either way.

PR #55, opened 2022-11-22
Clarify `fn` keyword's usage as argument

KorvinSzanto:
This PR removes the "preceding space" requirement from the `fn` keyword used in short closures in order to allow using short closures as arguments without needing to include a preceeding space.

```php
// Current spec:
foo( fn(int $x, int $y): int => $x + $y);

// After this PR:
foo(fn(int $x, int $y): int => $x + $y);


theodorejb:
Good catch. :+1:

samdark (approved):

PR #56, opened 2023-01-13
Assorted formatting and wording cleanup.

Crell:
1. Ensured that every example is preceded with language that makes it clear it is an example (or a bad example, in some cases).
2. Some whitespacing.
3. Collapsed some redundancy.

There should be no substantive changes in this PR.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -131,7 +131,7 @@ Short form of type keywords MUST be used i.e. `bool` instead of `boolean`,
 ### 2.6 Trailing commas
 
 Numerous PHP constructs allow a sequence of values to be separated by a comma,
-and the final item may have an optional comma.  Examples include array key/value pairs,
+and the final item may have an optional comma. Examples include array key/value pairs,
 function arguments, closure `use` statements, `match()` statement branches, etc.
 
 If that list is contained on a single line, then the last item MUST NOT have a trailing comma.
@@ -294,7 +294,7 @@ Any closing brace MUST NOT be followed by any comment or statement on the
 same line.
 
 When instantiating a new class, parentheses MUST always be present even when
-there are no arguments passed to the constructor.  For example:
+there are no arguments passed to the constructor. For example:
 
 ```php
 new Foo();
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -965,7 +964,8 @@ for ($i = 0; $i < 10; $i++) {
 Expressions in parentheses MAY be split across multiple lines, where each
 subsequent line is indented at least once. When doing so, the first expression
 MUST be on the next line. The closing parenthesis and opening brace MUST be
-placed together on their own line with one space between them.
+placed together on their own line with one space between them. The semicolons
+MUST be on the component the follow, not on the subsequent line.  For example:
```
theodorejb, line 968:
```suggestion
MUST be on the component they follow, not on the subsequent line.  For example:
```

theodorejb (changes requested):

spec.md
```diff
@@ -365,22 +364,11 @@ class ClassName extends ParentClass implements
...
-class ClassName
-{
-    use FirstTrait;
-}
-```
-
 Each individual trait that is imported into a class MUST be included
 one-per-line and each inclusion MUST have its own `use` import statement.
 
+The following is a correct example of trait `use`age.
```
samdark, line 370:
```suggestion
The following is a correct example of trait useage.
```

Mixing the code and the word itself looks extremely confusing.

samdark (approved):

spec.md
```diff
@@ -293,7 +294,7 @@ Any closing brace MUST NOT be followed by any comment or statement on the
 same line.
 
 When instantiating a new class, parentheses MUST always be present even when
-there are no arguments passed to the constructor.
+there are no arguments passed to the constructor.  For example:
```
KorvinSzanto, line 297: Can you update these to all use a single space or a newline before "For example"? I'd prefer sticking with single spaces everywhere.

spec.md
```diff
@@ -965,7 +964,8 @@ for ($i = 0; $i < 10; $i++) {
 Expressions in parentheses MAY be split across multiple lines, where each
 subsequent line is indented at least once. When doing so, the first expression
 MUST be on the next line. The closing parenthesis and opening brace MUST be
-placed together on their own line with one space between them.
+placed together on their own line with one space between them. The semicolons
+MUST be on the component they follow, not on the subsequent line.  For example:
```
KorvinSzanto, line 968: > The semicolons MUST be on the component they follow, not on the subsequent line

This is a new requirement, please remove this.

KorvinSzanto (changes requested):

KorvinSzanto (approved):

PR #62, opened 2023-02-26
Add a section about versioning to the meta document

### LINKED ISSUE #10: Versioning
I wanted to clearly define what "major" "minor" and "patch" mean in the context of this PER and how we will handle versioning for v1.0.0 and after. This is an important thing to nail down because we only need [core committee approval for major versions](https://www.php-fig.org/bylaws/per-workflow/#releases).

To make this easier to follow, let's split the available requirement keywords into two groups:
1. **"MUST"s:** `MUST`, `REQUIRED` or `SHALL`
2. **"SHOULD"s:** `SHOULD`, `RECOMMEND`, `MAY`, or `OPTIONAL`

- **Major:** Altered/added/removed `MUST`s *for already covered syntax* and altered/removed `SHOULD`s
- **Minor:** Added `SHOULD`s and added `MUST`s *for new syntax*
- **Patch:** Any small changes to text like typos, clarifications, and anything else that does not effect compatibility with any requirements

There are times when `SHOULD`s can be modified or even removed in backwards compatible ways and in those cases we may opt for a minor release rather than a major release, erring on the side of a major release (and therefore a core committee vote) whenever appropriate. **This is all up for discussion so if folks have any disagreements please add comments here**.

With all of that said [version 1.0.0 will be effectively a "patch" change from PSR-12](https://github.com/php-fig/per-coding-style/compare/8201676...master), all changes are done to align names and descriptions with this PER and no requirements have been altered/added/removed.

lpd-au:
Resolves #10

### What

This pull request adds a section to the meta document about versioning. It states that we use semantic versioning, details how semantic versioning is being applied to a text document rather than the usual software api, explains the rationale for that interpretation, gives a concrete example of that interpretation being applied to a hypothetical change and warns readers about the possibility of breaking changes in minor versions when using new PHP features before this PER standardises their style. Compared to the initial draft I shared in discord or the original comment in #10, it is less prescriptive about RFC 2119 keywords and more focused on the outcome for authors and projects.

### Why at all?

The [PER workflow bylaws](https://github.com/php-fig/fig-standards/blob/master/bylaws/003-per-workflow.md#releases) state that votes must be held for new releases that would be minor or major according to semantic versioning. Importantly, to my reading, it does not require that the releases themselves must be labelled according to semver; it would be equally valid to have release numbers that eg match php releases, provided votes are still held at the required times. Even if the bylaws did require that, they aren't an obvious place to look for spec readers. This addition makes it abundently clear we follow semver.

As written in the proposed section, semver is well defined when applied to software releases but has no common definition in other contexts. Even within the working group, our first instincts on how semver applies to a coding guidelines standard weren't a perfect match. In particular, the potential for breaking changes in minor versions if you use PHP features undocumented in the spec isn't necessarily intuitive. This section aims to convey to readers a high level understanding that is consistent with our current understanding, recognising we can't predict the low level nuance of every future change in advance.

### Why now?

Excluding 1.0 as a clone of PSR-12, this is effectively the first substantial release of any PER. How we handle releases is to me a _critical process issue_, both for ourselves and in establishing a precedent/example for future PERs. In my opinion, it is insufficient for an open github issue (especially with unresolved working group comments) to fulfil the role of documenting this critical process because:

1. It has low visibility to consumers of the spec.
2. The text is subject to change without notice.
3. The text has never been put to a working group vote.
4. The text is unenforcable.
5. It ignores best practice to document such decisions in the meta document.

Including this PR now, or an alternative wording, will:

1. Increase visibility to readers of the spec.
2. Formalise whatever guarantees we are prepared to offer consumers of the spec about minor and patch releases, and ensure the text cannot be changed unilaterally or without notice.
3. Demonstrate that the text represents the working group consensus, once 2.0 is accepted.
4. Bind us to either follow the text or follow proper processes to change the text, while working on 2.x/3.0 and beyond.
5. Ensure every substantial release of this PER has this critical process defined and set a positive example of best practice to future PERs.

### DIFF
--- a/meta.md
+++ b/meta.md
@@ -80,8 +80,7 @@ new code contributions processed by an automatic formatter follow (or disregard)
 existing code.
 
 Please note this backwards compatibility promise does not extend to projects that use new PHP syntax yet to be specified
-in this PER. In this circumstance, a minor version may introduce new requirements that conflict with syntax usage in 
-existing code.
+in this PER. In this circumstance, a minor version may introduce new requirements that are effectively breaking changes.
 
 It is ultimately determined whether a meaningful change may be included in a minor release on a case-by-case basis by 
 consensus. The addition of `MAY` or `OPTIONAL` requirements or the removal of requirements with non-optional RFC 2119 
--- a/meta.md
+++ b/meta.md
@@ -59,22 +59,35 @@ also not change anything stipulated in PSR-1 and PSR-12.
 
 ## 4. Versioning
 
-New releases of this PER are assigned version numbers in keeping with [semantic versioning](https://semver.org/). Semantic versioning is well defined when applied to software releases but has no common definition in other contexts. This PER applies the following meanings:
+New releases of this PER are assigned version numbers in keeping with [semantic versioning](https://semver.org/). 
+Semantic versioning is well defined when applied to software releases but has no common definition in other contexts. 
+This PER applies the following meanings:
 
 * **Patch** versions may contain:
-  * Any changes that do not alter the underlying requirements of the PER, such as fixing typos, adding clarifications or other modifications with no compatibility impact.
+  * Any changes that do not alter the underlying requirements of this PER, such as fixing typos, adding clarifications or 
+other modifications with no compatibility impact.
 * **Minor** versions may additionally:
-  * Add new `MAY` or `OPTIONAL` requirements.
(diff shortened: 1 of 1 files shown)


samdark (approved):

KorvinSzanto:
I had one suggestion but this looks good to me otherwise, thank you!

PR #72, opened 2023-04-21
Expand Properties and Constants on constants where it makes sense

samdark:


Crell:
Should we go further and apply the same to enum cases?

Crell:
That makes sense.  @samdark, can you expand the wording a bit?

Crell (approved):

Crell:
Valid.  OK, let's proceed here.

PR #73, opened 2023-05-04
Modified Keywords code sample contains parse errors

jrfnl:
* The `static` and `readonly` modifiers for properties are, at the time of writing, mutually exclusive, which makes the first property example a parse error.
* `final` properties are not a thing in PHP, which makes the second property example a parse error.

This commit fixes both code examples.

While I can appreciate that the code sample _intends_ to show an example where all possible modifier keywords + type declarations are used, I do not believe that showing this via code which isn't valid PHP to begin with, is helpful.

Introduced in #19.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -676,9 +676,9 @@ namespace Vendor\Package;
 
 abstract class ClassName
 {
-    protected static readonly string $foo;
+    protected static string $foo;
 
-    final protected int $beep;
+    private readonly int $beep;
 
     abstract protected function zim();
 


samdark (approved):

Crell:
A valid point, but to me that means we shouldn't just change the examples, but add more.  Like, "public static string $foo", "public readonly string $bar", etc.  Let's show several different valid permutations.  (Valid both by PHP and the PER.)  Can you update?

jrfnl:
> we shouldn't just change the examples, but add more. Like, "public static string $foo", "public readonly string $bar", etc. Let's show several different valid permutations. 

It already does and is also, why I changed `protected` in the second example to `private`.

The combined examples in that whole section now show all possible modifiers in use within valid PHP code.

PR #74, opened 2023-06-01
Consistently use php opening tag

ravage84:
Which is mandated by [3. Declare Statements, Namespace, and Import Statements](https://www.php-fig.org/per/coding-style/#3-declare-statements-namespace-and-import-statements), anyway.

Kinda "eat your own dog food".

Plus, it makes it even more clear in what file type context the code examples are.

Lastly, this also fixes thie following oddly looking (and unnecessary)  whitespace in [7.1 Short Closures](https://www.php-fig.org/per/coding-style/#71-short-closures):

grafik

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -141,6 +141,8 @@ If the list is split across multiple lines, then the last item MUST have a trail
 The following are examples of correct comma placement:
 
 ```php
+<?php
+
 function beep(string $a, string $b, string $c)
 {
     // ...
@@ -281,6 +283,8 @@ Block declare statements are allowed and MUST be formatted as below. Note positi
 braces and spacing:
 
 ```php
+<?php
+
 declare(ticks=1) {
     // some code
 }
@@ -297,6 +301,8 @@ When instantiating a new class, parentheses MUST always be present even when
 there are no arguments passed to the constructor. For example:
 
 ```php
+<?php
+
 new Foo();
 ```
 
@@ -305,6 +311,8 @@ then the body of the class SHOULD be abbreviated as `{}` and placed on the same 
 separated by a space. For example:
 
 ```php
+<?php
+
 class MyException extends \RuntimeException {}
 ```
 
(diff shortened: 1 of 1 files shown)


Crell:
I don't think this really applies here.  I almost never see the PHP opening tags in markdown code blocks.  The intent of the referenced section is that in a *file* you must start with the open tag, and use the long-tag version.  That doesn't really apply to code snippets.

Leaving this open to see if anyone else on the team disagrees, but I'm voting won't-fix here.

mbniebergall:
Including the opening PHP tag `<?php` is not standard for code snippets. See examples of no opening PHP tags at:
- https://www.php-fig.org/psr/psr-15/
- https://www.php-fig.org/psr/psr-17/
- https://www.php-fig.org/psr/psr-18/

Some PSrs _do_ include the opening tag, when conveying the context of the file, as seen in:
- https://www.php-fig.org/psr/psr-3/
- https://www.php-fig.org/psr/psr-20/#21-clockinterface

alekitto:
I agree that the opening tag should not be mandatory for code snippets.  
At the same time, I can see the inconsistency between the various PSRs: for example in PSR-3, the code block containing the `interpolate` function has an unnecessary opening tag, while the interface declarations (which are entire files) correctly have it.

I’m voting for wont-fix here, but probably we should define a general rule for code blocks/snippets

KorvinSzanto:
I don't see value in including the PHP tag in code snippets unless the snippet encompasses the entire file so I agree this isn't something we should do.

samdark:
I agree that there's no sense in adding it in these cases.

ravage84:
> (The unnecessary extra line in that one code block can be fixed in a separate PR.)

Separate PR: #80

ravage84:
> [...] At the same time, I can see the inconsistency between the various PSRs: for example in PSR-3, the code block containing the `interpolate` function has an unnecessary opening tag, while the interface declarations (which are entire files) correctly have it.

Actually, I disagree on that specific one because it could be the full code of this file., as well.
Adding a php opening tag implicitly hints at that & frames the whole code.

grafik

https://www.php-fig.org/psr/psr-3/#12-message

This would be a prime example where I would expect the php opening tag

ravage84:
While I like the consistency of having opening tags everywhere, I can see it being "too much", especially for one liners.

What do you think of the alternative in #81?

PR #75, opened 2023-06-01
Consistently format comments

ravage84:
- Capitalize beginning of all comments (regardless of keyword usage)
- Replace "..." with "{context} body"
- Use comment wording consistently & be specific for each use case
- No periods or ellipses at the end of comments

I'll admit, this is nitpicking but it helps to think less about the comments when reading the specification.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -639,7 +639,7 @@ name:
 ```php
 public function process(string $algorithm, ...$parts)
 {
-    // Method/function body
+    // Method body
 }
 ```
 
@@ -649,7 +649,7 @@ there MUST NOT be any space between the two of them:
 ```php
 public function process(string $algorithm, &...$parts)
 {
-    // Method/function body
+    // Method body
 }
 ```
 
@@ -928,7 +928,7 @@ of parentheses, spaces, and braces.
 <?php
 
 do {
-    // Structure body;
+    // Structure body
 } while ($expr);
 ```
 
@@ -941,7 +941,7 @@ always be at the beginning or at the end of the line, not a mix of both. For exa
 <?php
 
 do {
-    // Structure body;
+    // Structure body
 } while (
     $expr1
     && $expr2


Crell (approved):

spec.md
```diff
@@ -639,7 +639,7 @@ name:
 ```php
 public function process(string $algorithm, ...$parts)
 {
-    // processing
+    // Method/function body
```
KorvinSzanto, line 642: This is definitely a method and not a function since you can't use visibility modifiers with functions

spec.md
```diff
@@ -928,7 +928,7 @@ of parentheses, spaces, and braces.
 <?php
 
 do {
-    // structure body;
+    // Structure body;
```
KorvinSzanto, line 931: This PR leaves semicolons in some places and removes them from others, we should probably normalize those as well.

spec.md
```diff
@@ -987,7 +987,7 @@ parentheses, spaces, and braces.
 <?php
 
 foreach ($iterable as $key => $value) {
-    // foreach body
+    // Foreach body
```
KorvinSzanto, line 990: Capitalized `Foreach` is jarring to me. Why are we specifying the structure anyway when it's already clear from the preceding line of code? It'd be better for the comment to just be `// Body` or `// ...` in my opinion.

KorvinSzanto:
I agree these comments can be cleaned up and it might be good to do so. I'm not sure I agree with the choices here though, to me the `// ...` is the gold standard as opposed to something overly verbose like `// Here lies a placeholder for class body content` as the comments aren't really meant to be read, they are meant to act as a stand-in for code that would otherwise have to be dummy code.

samdark (approved):
👍 except @KorvinSzanto comments

spec.md
ravage84: Agreed. Laziness on my part.

Fixed.

spec.md
ravage84: Agreed. Oversight on my part.

Fixed.

spec.md
ravage84: I guess this comes down to personal preference. I have no strong feelings either way, I simply prefer consistency because inconsistency is jarring to me.

I created an alterantive to this PR: #79

ravage84:
I created an alternative to this PR that replaces all strucutre body comments with ellipsis: https://github.com/php-fig/per-coding-style/pull/79

Crell:
Merged #79 instead, so closing this one.

PR #77, opened 2023-06-01
Do not mention PHP support

ravage84:
This was the only spot in which the PHP support & version was mentioned. Either all statements should have or none.

But it superfluous since the section "[Previous language versions](https://www.php-fig.org/per/coding-style/#previous-language-versions)" already covers it:

> Throughout this document, any instructions MAY be ignored if they do not exist in versions of PHP supported by your project.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -443,8 +443,7 @@ class Talker
 
 Visibility MUST be declared on all properties.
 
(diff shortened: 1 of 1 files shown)


Crell:
Makes sense, thanks.

PR #79, opened 2023-06-05
Replace all structure body comments with simple ellipsis

ravage84:
- Capitalize beginning of all comments
- No periods at the end of comments

As an alternative to #75.


Crell (approved):

Crell:
An informal poll in the WG Discord overwhelmingly favored this version, so we'll go with it.  Thanks.

PR #81, opened 2023-06-05
Only add PHP opening tag when the following rule of thumb are met

ravage84:
While I like the consistency of having opening tags everywhere (as in #74), I can see it being "too much", especially for one liners.

What do you think of the following rule of thumb proposal?

Add PHP opening tags when all requirements are met:

- Tne [mandatory class namespace](https://www.php-fig.org/psr/psr-1/#3-namespace-and-class-names) is defined or no class or interface respectively is defined
- All necessary classes & functions are imported
- All necessary variables & constants are defined
- All necessary structures are visible or hinted at with a comment
- All statements and/or structures are related to each other
- Or the PHP opening tag is essential to the code example

Do not add PHP opening tags when:

- Only one or many unrelated statements are shown
- Only one or many unrelated functions are shown
- Only one or many unrelated methods are shown
- Only one or many unrelated classes or interfaces respectively are shown
- Only one or many unrelated other structures, such as an if-clause or match expression, are shown

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -234,8 +234,6 @@ When using compound namespaces, there MUST NOT be more than two sub-namespaces w
 That is, the following is allowed:
 
 ```php
-<?php
-
 use Vendor\Package\SomeNamespace\{
     SubnamespaceOne\ClassA,
     SubnamespaceOne\ClassB,
@@ -247,8 +245,6 @@ use Vendor\Package\SomeNamespace\{
 And the following would not be allowed:
 
 ```php
-<?php
-
 use Vendor\Package\SomeNamespace\{
     // This has too many namespace segments to be in a group.
     SubnamespaceOne\AnotherNamespace\ClassA,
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -234,8 +234,6 @@ When using compound namespaces, there MUST NOT be more than two sub-namespaces w
 That is, the following is allowed:
 
 ```php
-<?php
-
```
Crell, line 238: I'd leave this one.  We're talking about the `use` statements, which happen at a very specific place in the file, so I think it's fine to leave here.

vjik:
Suggest shorten the list when need add PHP opening tag to several items:
- block contain code with several sections (for example, class declaration + use) that can be placed whole in file;
- the PHP opening tag is essential to the code example.

In otherwise cases `<?php` looks overhead.

Crell:
Looks like nothing is happening here.

PR #90, opened 2024-07-21
Add missing space in example for anonymous class declaration

rikvdh:
According to the specification:

> Anonymous Classes MUST follow the same guidelines and principles as closures in the above section.

Therefore a space must be between the class keyword and the opening parenthesis, as with closures.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1242,7 +1242,7 @@ $instance = new class extends \Foo implements \HandleableInterface {
 
 // Brace on the next line
 // Constructor arguments
-$instance = new class($a) extends \Foo implements
+$instance = new class ($a) extends \Foo implements
     \ArrayAccess,
     \Countable,
     \Serializable


samdark:
Good catch, but I think the text should be fixed and not the example.

jrfnl:
> Good catch, but I think the text should be fixed and not the example.

Shouldn't the example be fixed for the _current_ version of PER ? Only for the text (and example) to be changed/fixed in the _next_ version of PER ?

jrfnl:
As a side-note: coding standards should be about consistency, "fixing" the text to make the spacing requirements for closures and anonymous classes different makes things _in_consistent.

Crell:
@jrfnl Off hand, do you know what CS and CSFixer do here?  Do they follow the text or the example?

jrfnl:
> @jrfnl Off hand, do you know what CS and CSFixer do here? Do they follow the text or the example?

I can't speak for CS-Fixer, but PHPCS follows the text of the PSR **_to the letter_** and only falls back to the examples when the text is unclear or states "as per the example ...".

For this particular case, I just happen to know that both PHPCS as well as CS-Fixer enforce one space. See https://github.com/squizlabs/PHP_CodeSniffer/issues/3200 for a previous discussion about this (and confirmation).

Crell (approved):

samdark (approved):

PR #93, opened 2024-08-12
Add ternary operator on multi-line

JesusValera:
## 📚 Description

Some people write ternary operators as follows:

```php
$foo = true ?
    'foo' :
    'bar';
```

In the previous snippet is kind of easy, but when instead of `'foo'`, it is a long line calling method, etc, it becomes complex.

My proposal is to add the following rule into the PER file to avoid people doing the previous snippet, making the code easier to track when ternary operators are used to improve the legibility; eg:
```php
$foo = true
    ? 'foo'
    : 'bar';
```

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1068,18 +1068,21 @@ $variable = $foo ?: 'bar';
 
 ### 6.4. Operator's placement
 
-When a statement that includes an operator must be split into multiple lines,
-the operator SHOULD be placed at the beginning of the new line.
+A statement that includes an operator MAY be split across multiple lines, where
+each subsequent line is indented once. When doing so, the operator MUST be
+placed at the beginning of the new line; ternaries MUST occupy 3 lines, never 2.
+
+For example:
 
 ```php
 <?php
 
-$variable1 = $possibleNullableExpr
-    ?? 'fallback';
-
-$variable2 = $ternaryOperatorExpr
+$variable1 = $ternaryOperatorExpr
     ? 'fizz'
     : 'buzz';
+
+$variable2 = $possibleNullableExpr
+    ?? 'fallback';
 
 $variable3 = $elvisExpr
     ?: 'qix';
--- a/spec.md
+++ b/spec.md
@@ -1057,10 +1057,6 @@ and `:` characters:
 
 ```php
 $variable = $foo ? 'foo' : 'bar';
-
-$variableMultiLine = $foo
-    ? 'foo'
-    : 'bar';
 ```
 
 When the middle operand of the conditional operator is omitted, the operator
@@ -1068,6 +1064,25 @@ MUST follow the same style rules as other binary [comparison][] operators:
 
 ```php
 $variable = $foo ?: 'bar';
+```
+
+### 6.4. Operator's placement
+
+When a statement that includes an operator must be split into multiple lines,
+the operator SHOULD be placed at the beginning of the new line.
+
+```php
+<?php
+
+$variable1 = $possibleNullableExpr
+    ?? 'fallback';
+
+$variable2 = $ternaryOperatorExpr
+    ? 'fizz'
+    : 'buzz';
+
+$variable3 = $elvisExpr
+    ?: 'qix';
 ```
 
 ## 7. Closures
--- a/spec.md
+++ b/spec.md
@@ -1057,6 +1057,10 @@ and `:` characters:
 
 ```php
 $variable = $foo ? 'foo' : 'bar';
+
+$variableMultiLine = $foo
+    ? 'foo'
+    : 'bar';
 ```
 
 When the middle operand of the conditional operator is omitted, the operator


Crell:
We probably need some more explanatory text on there, but as a fan of multi-line ternaries, I support this addition.

Anyone else from the WG want to weigh in?

JesusValera:
I was thinking, maybe it makes sense to also add this rule for the Elvis/null coalescing operators?

```php
// Not this
$foo = 'foo' ??
    'bar';

$foo = 'foo' ?:
    'bar';

// But this
$foo = 'foo'
    ?? 'bar';

$foo = 'foo'
    ?: 'bar';
```

I could add the modifications in this (or another PR), let me know what you think and if you have another suggestion 😄

samdark (approved):

Crell:
Let's make that a separate PR so it can have its own discussion.

Jean85:
Can I say that I totally prefer the opposite style, with the operators at the start of the line? 
```php
$foo = $bar
    ? 'bar'
    : 'foo';
```

Wouldn't it be more consistent with how binary operators are considered?

JesusValera:
> Can I say that I totally prefer the opposite style, with the operators at the start of the line?
> 
> ```
> $foo = $bar
>     ? 'bar'
>     : 'foo';
> ```
> 
> Wouldn't it be more consistent with how binary operators are considered?

Sorry, I do not understand your statement - do you mean you prefer them at the beginning or the end?
The proposal is to have the operators at the beginning 😄

rhertogh:
@Jean85 @JesusValera I initially had the same misunderstanding based on the opening post (before looking at the file changes.)
Perhaps it would be good to add the proposed format to the opening post?

Jean85:
:+1: on what @rhertogh said :smile: I got confused too!

JesusValera:
I just pushed a new commit.

Initially, I was unsure if the content of this PR belonged to the `2.3 Lines`, the `6.3. Ternary operators` or a new `6.4. Operator's placement` section.
I decided to create the new section as the proposal involves not only the `lines` related section but also the operator behaviours when multi-lines, because of that, probably a new section is the best choice.

Let me know what you think, otherwise, I could change the code again :octocat:

spec.md
```diff
@@ -1066,6 +1066,25 @@ MUST follow the same style rules as other binary [comparison][] operators:
 $variable = $foo ?: 'bar';
 ```
 
+### 6.4. Operator's placement
+
+When a statement that includes an operator must be split into multiple lines,
+the operator SHOULD be placed at the beginning of the new line.
```
Crell, line 1072: This should also mention the indentation, explicitly.

I am also tempted to make this a MUST, but would like input from the rest of the WG.

JesusValera: @Crell not sure if I understand what you mean with `mention explicitly the identation`.
Do you mean something like?

```
When a statement that includes an operator must be split into multiple lines,
the operator SHOULD be placed at the beginning of the new line preceded
by 4 spaces.
```

Regarding the `SHOULD` or `MUST`, I think `MUST` could work better here 🤔

KorvinSzanto: Consider this example for lists of implements:

> Lists of implements and, in the case of interfaces, extends MAY be split across multiple lines, where each subsequent line is indented once. When doing so, the first item in the list MUST be on the next line, and there MUST be only one interface per line. For example:

I think we probably want most of these things for multiline ternaries too:
- "MAY be split across multiple lines" It's important we allow this explicitly with an RFC 2119 word
- "where each subsequent line is indented once" Require the indentation
- "MUST be only one interface per line" When split on multiple lines, ternaries must occupy at least 3 lines, never 2.

So the golden path is allowed explicitly and the following ends up disallowed explicitly:
```php
$foo = $baz
? 'fizz'
: 'buzz';

$foo = $baz
    ? 'fizz' : 'buzz';

$foo = $baz
       ? 'fizz' 
       : 'buzz';

Comment (JesusValera) [reply spec.md]: It should be better now. If you still think something could be improved, let me know 😃

**Review by Chemaclass — APPROVED** [review]

**Review by Crell — APPROVED** [review]

File: spec.md [hunk 2]
```diff
@@ -1066,6 +1066,28 @@ MUST follow the same style rules as other binary [comparison][] operators:
 $variable = $foo ?: 'bar';
 ```
 
+### 6.4. Operator's placement
+
+A statement that includes an operator MAY be split across multiple lines, where
+each subsequent line is indented once. When doing so, the operator MUST be
+placed at the beginning of the new line; ternaries MUST occupy 3 lines, never 2.
```
TimWolla, line 1073: This is contradicting 5.2, which says:

> Boolean operators between conditions MUST always be at the beginning or at the end of the line, not a mix of both.

JesusValera: That's right! And this line is repeated 4 times. It would be better to drop the "`or at the end of the line`" part.

What do you think?

Crell: Hm.  Yes, please file a new PR to remove that clause.

JesusValera: https://github.com/php-fig/per-coding-style/pull/103

PR #99, opened 2024-10-15
Document how to use aviz

### LINKED ISSUE #98: New PHP 8.4 Syntax: Asymmetric visibility
https://wiki.php.net/rfc/asymmetric-visibility-v2

```php
class Foo
{
    PRIVATE(set) string $bar1 = '';
    public private(SET) string $bar2 = '';
    private(set) public string $bar3 = '';

    public function __construct(
        public protected(SET) string $bar4 = ''
    ) {}
}
```

See this working here: https://3v4l.org/JDXuq/rfc#vgit.master_jit

Crell:
Resolves #98.

This seems the obvious way to do it.  I debated the RECOMMENDED part.  I'm open to changing that if other WG members feel strongly about it.

WG members (only): Please :+1: this thread to indicate your approval, or comment to the contrary.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -659,12 +659,13 @@ engine and language handles them. When present, they MUST be in the following or
 
 * Inheritance modifier: `abstract` or `final`
 * Visibility modifier: `public`, `protected`, or `private`
+* Set-visibility modifier: `public(set)`, `protected(set)`, or `private(set)`
 * Scope modifier: `static`
 * Mutation modifier: `readonly`
 * Type declaration
 * Name
 
-All keywords MUST be on a single line, and MUST be separated by a single space.
+All keywords MUST be on a single line, and MUST be separated by a single space.  All keywords MUST be all lower-case.  When using a set-visibility on a public-read property, omitting the `public` keyword is RECOMMENDED but NOT REQUIRED.
 
 The following is a correct example of modifier keyword usage:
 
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -659,12 +659,13 @@ engine and language handles them. When present, they MUST be in the following or
...
 * Inheritance modifier: `abstract` or `final`
 * Visibility modifier: `public`, `protected`, or `private`
+* Set-visibility modifier: `public(set)`, `protected(set)`, or `private(set)`
 * Scope modifier: `static`
 * Mutation modifier: `readonly`
 * Type declaration
 * Name
 
-All keywords MUST be on a single line, and MUST be separated by a single space.
+All keywords MUST be on a single line, and MUST be separated by a single space.  All keywords MUST be all lower-case.  When using a set-visibility on a public-read property, omitting the `public` keyword is RECOMMENDED but NOT REQUIRED.
```
KorvinSzanto, line 668: `NOT REQUIRED` is not an RFC 2119 keyword so it should be lowercase and probably omitted.
Though I would opt for a `MAY` rather than recommend:

```suggestion
All keywords MUST be on a single line, and MUST be separated by a single space.  All keywords MUST be all lower-case.  The `public` keyword MAY be omitted when using a set-visibility on a public-read property.
```

Though frankly I probably wouldn't advocate omitting the public, the same way I wouldn't omit public when writing public properties today.

KorvinSzanto (approved):
LGTM

PR #100, opened 2024-10-15
Clarify that php open tags must be lowercase

### LINKED ISSUE #91: Opening PHP tag in uppercase
None of the PSRs for Coding Standards, nor PER-CS, explicitly indicate whether the opening tag for PHP should be in lowercase, though this is very much implied in PSR-1 - see https://www.php-fig.org/psr/psr-1/#1-overview

It probably should be explicitly stated in the next version of PER-CS.

Also, it should be noted that php-cs-fixer does force the opening tag to be in lowercase, whilst phpcs does not.

Crell:
Resolves #91.

WG members (only): Please 👍 this thread to indicate your approval, or comment to the contrary.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -192,7 +192,7 @@ PHP.
 
 When the opening `<?php` tag is on the first line of the file, it MUST be on its
 own line with no other statements unless it is a file containing markup outside of PHP
-opening and closing tags.
+opening and closing tags.  The `<?php` tag MUST always be lower-case.
 
 Import statements MUST never begin with a leading backslash as they
 must always be fully qualified.


samdark (approved):

lpd-au:
Small nitpick, the spec currently has 2 uses of `lower case`, 0 uses of `lower-case` and 0 uses of `lowercase`.

spec.md
```diff
@@ -192,7 +192,7 @@ PHP.
 
 When the opening `<?php` tag is on the first line of the file, it MUST be on its
 own line with no other statements unless it is a file containing markup outside of PHP
-opening and closing tags.
+opening and closing tags.  The `<?php` tag MUST always be lower-case.
```
Crell, line 195:
```suggestion
opening and closing tags.  The `<?php` tag MUST always be lower case.
```

English is dumb.

PR #108, opened 2025-01-04
Specify hooks formatting

### LINKED ISSUE #88: New PHP 8.4 Syntax: Property Hooks
[Property hooks](https://wiki.php.net/rfc/property-hooks) are landing in 8.4, the next version of PER-CS should cover them:

```php
class Foo
{
    public function __construct(private User $user) {}

    public string $userName {
        get {
            return $this->user->userName;
        }
    }
}
```

Crell:
So we have something concrete to chew on.

Having used hooks on a project for the last few months, this feels about right to me.  I could see arguments for other restrictions on when to use one-big-line hooks, but they do have their place.  (Passthroughs, for instance.)

Resolves #88

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -899,7 +899,7 @@ class Example
 
 ## 4.10 Interface properties
 
-Abstract roperties may be defined in interfaces or abstract classes, but are required to
+Abstract properties may be defined in interfaces or abstract classes, but are required to
 specify if they must support `get` operations, `set` operations, or both.
 
 * The operation block MUST be on the same line as the property.
--- a/spec.md
+++ b/spec.md
@@ -772,7 +772,6 @@ A function or method may be referenced in a way that creates a closure out of it
 
 If so, the `...` MUST NOT include any whitespace before or after. That is, the correct format is `foo(...)`.
 
-
 ### 4.9 Property Hooks
 
 Object properties may also include hooks, which have a number of syntactic options.
@@ -897,10 +896,27 @@ class Example
 }
 ```
 
-## 4.10 Interface properties
+## 4.10 Interface and abstract properties
 
 Abstract properties may be defined in interfaces or abstract classes, but are required to
-specify if they must support `get` operations, `set` operations, or both.
+specify if they must support `get` operations, `set` operations, or both.  In the case
+of abstract classes, they MAY include a body for one or another hook.
+
+If there is a body for any hook, then the entire hook block MUST follow
+the same rules as for defined hooks above.  The only difference is that
+a hook that has no body specified have a single semicolon after the hook
+keyword, with no space before it.
+
+```php
+abstract class Example {
+    abstract public string $name {
+        get => ucfirst($this->name);
+        set;
+    }
+}
+```
+
+If there is no body for either hook, then the following rules apply:
 
 * The operation block MUST be on the same line as the property.
 * There MUST be a single space between the property name and the operation block `{}`.
--- a/spec.md
+++ b/spec.md
@@ -848,6 +848,7 @@ Additionally, if the following criteria are met:
 
 * There is only one hook implementation.
 * That hook uses the short-hook syntax.
+* That hook expression does not contain any wrapping.
 
 Then the hook MAY be listed entirely inline.  In that case,
 
--- a/spec.md
+++ b/spec.md
@@ -479,6 +479,109 @@ class ClassName
 
 ### 4.4 Property Hooks
 
+Object properties may also include hooks, which have a number of syntactic options.
+
+If there is only one hook implementation, and that hook implementation itself uses
+the short-hook syntax, then the hook declaration MAY be listed entirely inline.  The hook
+name MUST be separated from the opening brace and the arrow operator by a single
+space, and the semicolon ending of the hook MUST be separated from the cosing brace
+by a single space.  For example:
+
+```php
+class Example
+{
+    public string $myName { get => __CLASS__; }
+
+    public string $newName { set => ucfirst($value); }
+}
+```
+
+If those two criteria are not met, then the hook block MUST be spread across multiple lines:
+
+```php
+class Example
+{
+    public string $myName {
+        get => __CLASS__;
+    }
+
+    public string $newName {
+        set => ucfirst($value);
+    }
+}
+```
+
+When using the multi-line form, the opening brace MUST be on the same line as the property and
+the closing brace MUST be on its own line, with the body indented one level.
+
+The `=>` operator MUST have a single space on either side.
+
+If multiple hooks are declared, they MUST be separated by at least a single line break.  They
+MAY be separated by an additional blank line to aid readability.
+
+If the property includes a default value, there MUST be a single space between the default value
+and the opening brace.
+
+```php
+class Example
+{
+    public string $newName = 'Me' {
+        set => ucfirst($value);
+    }
+}
+```
+
+If a hook is using its multi-line form, the opening brace MUST be on the same line as the hook
+name, and the closing brace MUST be on its own line.  The body MUST be indented one level.
+
+```php
+class Example
+{
+    public string $newName = 'Me' {
+        set {
+            if (strlen($value) < 3) {
+                throw new \Exception('Too short');
+            }
+            $this->newName = ucfirst($value);
+        }
+    }
+}
+```
+
+For a `set` hook, if the argument name and type do not need to be redefined, then they MAY be omitted.
+
+Property hooks MAY also be defined in constructor-promoted properties.  However, they
+MUST be only a single hook, with a short-syntax body, defined on a single line as above.
+If those criteria are not met, then the promoted property MUST NOT have any hooks defined
+inline, and SHOULD instead be defined as normal property and not promoted.
+
+```php
+class Example
+{
+    public function __construct(
+        public string $name { set => ucfirst($value; }
+    ) {}
+}
+```
+
+The following is ***not allowed*** due to the hook being too complex:
+
+```php
+class Example
+{
+    public function __construct(
+        public string $name {
+            set {
+                if (strlen($value) < 3) {
+                    throw new \Exception('Too short');
+                }
+                $this->newName = ucfirst($value);
+            }
+        }
+    ) {}
+}
+```
+
 ### 4.5 Methods and Functions
 
 Visibility MUST be declared on all methods.
--- a/spec.md
+++ b/spec.md
@@ -828,7 +828,8 @@ For a `set` hook, if the argument name and type do not need to be redefined, the
 If a hook consists of a single expression, then PHP allows it to be shortened using `=>`.  In that case:
 
 * There MUST be a single space on either side of the `=>` symbol.
-* The body MUST be on the same line as the hook declaration and MUST NOT be wrapped. If wrapping is necessary to keep the line within the line length or desired for other reasons the short-hook syntax MUST NOT be used.
+* The body MUST begin on the same line as the hook name and `=>`.
+* Wrapping is allowed if the expression used allows for wrapping, using the rules defined elsewhere in this document.
 
 ```php
 class Example
--- a/spec.md
+++ b/spec.md
@@ -477,7 +477,9 @@ class ClassName
 }
 ```
 
-### 4.4 Methods and Functions
+### 4.4 Property Hooks
+
+### 4.5 Methods and Functions
 
 Visibility MUST be declared on all methods.
 
@@ -542,7 +544,7 @@ class Point
 }
 ```
 
-### 4.5 Method and Function Parameters
+### 4.6 Method and Function Parameters
 
 In the argument list, there MUST NOT be a space before each comma, and there
 MUST be one space after each comma.
@@ -660,7 +662,7 @@ public function process(string $algorithm, &...$parts)
 }
 ```
 
-### 4.6 Modifier Keywords
+### 4.7 Modifier Keywords
 
 Classes, properties, and methods have numerous keyword modifiers that alter how the
 engine and language handles them. When present, they MUST be in the following order:
@@ -701,7 +703,7 @@ readonly class ValueObject
 }
 ```
 
-### 4.7 Method and Function Calls
+### 4.8 Method and Function Calls
 
 When making a method or function call, there MUST NOT be a space between the
 method or function name and the opening parenthesis, there MUST NOT be a space
@@ -766,7 +768,7 @@ $someInstance
     ->run();
 ```
 
-### 4.8 Function Callable References
+### 4.9 Function Callable References
 
 A function or method may be referenced in a way that creates a closure out of it, by providing `...` in place of arguments.
 
--- a/spec.md
+++ b/spec.md
@@ -897,6 +897,30 @@ class Example
 }
 ```
 
+## 4.10 Interface properties
+
+Abstract roperties may be defined in interfaces or abstract classes, but are required to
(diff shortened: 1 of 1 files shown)


samdark (approved):

spec.md
```diff
@@ -477,7 +477,112 @@ class ClassName
...
 
-### 4.4 Methods and Functions
+### 4.4 Property Hooks
+
+Object properties may also include hooks, which have a number of syntactic options.
+
+If there is only one hook implementation, and that hook implementation itself uses
+the short-hook syntax, then the hook declaration MAY be listed entirely inline.  The hook
+name MUST be separated from the opening brace and the arrow operator by a single
+space, and the semicolon ending of the hook MUST be separated from the cosing brace
```
TimWolla, line 487:
```suggestion
space, and the semicolon ending of the hook MUST be separated from the closing brace
```

spec.md
```diff
@@ -477,7 +477,112 @@ class ClassName
...
+    public string $myName { get => __CLASS__; }
+
+    public string $newName { set => ucfirst($value); }
+}
+```
+
+If those two criteria are not met, then the hook block MUST be spread across multiple lines:
+
+```php
+class Example
```
TimWolla, line 502: This example does not match the preceding paragraph, because the criteria are still met.

spec.md
```diff
@@ -477,7 +477,112 @@ class ClassName
...
+    }
+}
+```
+
+For a `set` hook, if the argument name and type do not need to be redefined, then they MAY be omitted.
+
+Property hooks MAY also be defined in constructor-promoted properties.  However, they
+MUST be only a single hook, with a short-syntax body, defined on a single line as above.
+If those criteria are not met, then the promoted property MUST NOT have any hooks defined
+inline, and SHOULD instead be defined as normal property and not promoted.
```
TimWolla, line 556: This SHOULD doesn't make sense here, because there is no other option than to use a normal property.

spec.md
```diff
@@ -477,7 +477,112 @@ class ClassName
...
+
+```php
+class Example
+{
+    public function __construct(
+        public string $name { set => ucfirst($value; }
```
TimWolla, line 562:
```suggestion
        public string $name { set => ucfirst($value); }
```

TimWolla (commented):
Apart from the inline remarks, as I reader I would expect the “expanded” syntax to be defined first: That's the only syntax that works in every case. The short syntax then is a legal exception that MAY be used.

Currently it reads quite awkward, because the definition starts with the exception “If […]”, without properly explaining the “else” part.

-----------

Also as mentioned in the issue, I disagree with the brace placement because of the inconsistency with methods. The second example given by @edorian in https://github.com/php-fig/per-coding-style/issues/88#issuecomment-2416775499 looks like the best compromise to me.

spec.md
```diff
@@ -477,7 +477,112 @@ class ClassName
 }
 ```
 
-### 4.4 Methods and Functions
+### 4.4 Property Hooks
```
KorvinSzanto, line 480: IMO we should add this to the end of 4 rather than displacing everything else in 4. Ideally one could refer to `PER-CS 4.5` and expect it to remain the same across _at least_ minor versions.

samdark: Agree.

Crell:
* Rewrote the hooks section to go from longer form to shorter form.
* Used more bullet points.
* Moved the hooks section later.
* Added a section on interface properties, too.

Thoughts now?

Crell:
As for the brace spacing (I deliberately left that for the moment), here's some examples from a project I'm working on now in 8.4 that makes heavy use of hooks.

```php
    public private(set) string $logicalPath {
        get => $this->logicalPath ??= $this->parsedFolder->logicalPath;
    }

    private PageSet $children {
        get => $this->children ??= new BasicPageSet($this->pageTree->folderAllPages($this->logicalPath));
    }
```

Fun fact, I don't think I'm using the long-form version at all yet. :smile: Or the set hook.

I find the above approach completely readable and logical.  If we make the outer braces own-line:

```php
    public private(set) string $logicalPath
   {
        get => $this->logicalPath ??= $this->parsedFolder->logicalPath;
    }

    private PageSet $children
    {
        get => $this->children ??= new BasicPageSet($this->pageTree->folderAllPages($this->logicalPath));
    }
```

That doesn't actually feel any more readable to me.  It just feels like more scrolling, especially if I had more than two properties on this class.  (I have another that has 10, but they're all inlineable.)

Crell:
Actually here's a better example from the same project:

```php
class PageRecord
{
    // @todo More robust than this.
    public int $title {
        get => $this->values(__PROPERTY__)[0];
    }

    public int $order {
        get => \max($this->values(__PROPERTY__));
    }

    public bool $hidden {
        get => array_all($this->values(__PROPERTY__), static fn($x): bool => (bool)$x);
    }

    public bool $routable {
        get => array_any($this->values(__PROPERTY__), static fn($x): bool => (bool)$x);
    }

    public bool $isFolder {
        get => array_any($this->values(__PROPERTY__), static fn($x): bool => (bool)$x);
    }

    public \DateTimeImmutable $publishDate {
        get => \max($this->values(__PROPERTY__));
    }

    public \DateTimeImmutable $lastModifiedDate {
        get => \max($this->values(__PROPERTY__));
    }

    public string $pathName {
        get => substr($this->logicalPath, strrpos($this->logicalPath, '/') + 1);
    }

    public array $tags {
        get => array_values(array_unique(array_merge(...$this->values(__PROPERTY__))));
    }

    public function __construct(
        public string $logicalPath,
        public string $folder,
        public array $files,
    ) {}

    private function values(string $property): array
    {
        return array_column($this->files, $property);
    }
}
```

Adding another line break in there gives me no value, it just makes me scroll more.

spec.md
```diff
@@ -772,6 +772,155 @@ A function or method may be referenced in a way that creates a closure out of it
...
+```
+
+Property hooks also support multiple short-hook variations.
+
+For a `set` hook, if the argument name and type do not need to be redefined, then they MAY be omitted.
+
+If a hook consists of a single expression, then PHP allows it to be shortened using `=>`.  In that case:
+
+* There MUST be a single space on either side of the `=>` symbol.
+* The body MUST be on the same line as the hook declaration, unless it gets prohibitively long.  If it gets prohibitively long, the developer SHOULD consider not using the short-hook syntax.
```
TimWolla, line 832: This entire line is not sufficiently actionable. Combining a MUST with an unless doesn't make sense.

We already have a character limit per line, perhaps we can refer to that here?

> The body MUST be on the same line as the hook declaration and MUST NOT be wrapped. If wrapping is necessary to keep the line within the line length or desired for other reasons the short-hook syntax MUST NOT be used.

TimWolla (commented):
The new order is much better, thank you.

spec.md
```diff
@@ -772,6 +772,155 @@ A function or method may be referenced in a way that creates a closure out of it
...
+    ) {}
+}
+```
+
+## 4.10 Interface properties
+
+Abstract roperties may be defined in interfaces or abstract classes, but are required to
+specify if they must support `get` operations, `set` operations, or both.
+
+* The operation block MUST be on the same line as the property.
```
TimWolla, line 905: I'm not sure if I agree with the MUST here, because that effectively requires the user to use the short-form and thus is inconsistent with hooks themselves.

I find it reasonable to write the following in an interface:

```
public string $both {
    get;
    set;
}
```

And then shortening that to a single line MAY be done (consistently with hooks).

spec.md
```diff
@@ -772,6 +772,155 @@ A function or method may be referenced in a way that creates a closure out of it
...
+* If multiple hooks are declared, they MUST be separated by at least a single line break.  They
+  MAY be separated by an additional blank line to aid readability.
+
+For example:
+
+```php
+class Example
+{
+    public string $newName = 'Me' {
+        set {
```
TimWolla, line 796:
```suggestion
        set(string $value) {
```

To actually show the long-long form in one of the cases.

spec.md
Crell: Can you make an edit for that so I can merge it?

spec.md
Crell: There's actually a subtle difference here.  For interfaces, there can never be a body to a hook, so I see no point in the extra lines.  (If Levi ever re-proposes interface default methods that may change, but we'll cross that bridge when we get to it.)

For abstract classes, though, one could specify one hook as abstract and one as predefined.  In that case, multi-lining is probably better.

We should probably address those separately.

spec.md
```diff
@@ -772,6 +772,155 @@ A function or method may be referenced in a way that creates a closure out of it
...
+                $this->newName = ucfirst($value);
+            }
+        }
+    ) {}
+}
+```
+
+## 4.10 Interface properties
+
+Abstract roperties may be defined in interfaces or abstract classes, but are required to
```
KorvinSzanto, line 902:
```suggestion
Abstract properties may be defined in interfaces or abstract classes, but are required to
```

spec.md
```diff
@@ -772,6 +772,155 @@ A function or method may be referenced in a way that creates a closure out of it
...
+specify if they must support `get` operations, `set` operations, or both.
```
KorvinSzanto, line 903: I've opened https://github.com/php-fig/per-coding-style/pull/111 so that lowercase `must`s aren't ambiguous. I don't think this line would actually cause any confusion, but it spurred me to submit the PR.

spec.md
```diff
@@ -772,6 +772,155 @@ A function or method may be referenced in a way that creates a closure out of it
...
+```
+
+Property hooks also support multiple short-hook variations.
+
+For a `set` hook, if the argument name and type do not need to be redefined, then they MAY be omitted.
+
+If a hook consists of a single expression, then PHP allows it to be shortened using `=>`.  In that case:
+
+* There MUST be a single space on either side of the `=>` symbol.
+* The body MUST be on the same line as the hook declaration, unless it gets prohibitively long.  If it gets prohibitively long, the developer SHOULD consider not using the short-hook syntax.
```
TimWolla, line 832:
```suggestion
* The body MUST be on the same line as the hook declaration and MUST NOT be wrapped. If wrapping is necessary to keep the line within the line length or desired for other reasons the short-hook syntax MUST NOT be used.
```

spec.md
TimWolla: > so I see no point in the extra lines

The point is _allowing_ for an uniform formatting of similar things. In this case: A property declaration with a list of “operations”.

I also find it much easier to scan the code vertically rather than horizontally, because the of the established pattern of “a line-break indicates that something new begins, with indentation indicating the relationship to the previous line”.

lkrms:
It's great to see progress being made on this--not least because it seems to line up reasonably well with my approach to property hooks in [pretty-php](https://github.com/lkrms/pretty-php) (hardly surprising given how PSR-12- and PER-influenced this little ~~obsession~~ side project has become).

I just formatted the most recent code listings from this PR with `pretty-php` and there are some differences around what it accepts on a single line (this PR is more accepting so far 😊), but I can see there are ongoing discussions about that and will weigh in separately if I have something useful to add.

Otherwise, my only concern so far relates to the spacing of the long form of hooks, e.g. after `set` here:

```php
class Example
{
    public string $newName = 'Me' {
        set(string $value) {
            //
        }
    }
}
```

My approach has been to format the short form like an arrow function, and the long form like an anonymous function (as opposed to a function named `get` or `set`). Their equivalence feels familiar on this basis, and I think adding a space after `set` improves readability while maintaining consistency with existing closure formatting requirements:

```php
class Example
{
    public string $newName = 'Me' {
        set (string $value) {
            //
        }
    }
}
```

I do apologise if this has already been covered--I may not have expanded every thread--but I didn't see it mentioned.

---

(And with that, I should probably ask if there is somewhere I should go to introduce myself properly? Long-time stalker, first time talker, looking to engage and connect more as part of getting out of an extended period of isolation and health difficulties, of which `pretty-php` is a somewhat of a by-product...)

Crell:
I adjusted the language for abstract properties, to handle those with bodies separately from those with bodies.  I think that covers it.

Crell:
@lkrms Welcome!  I don't think anyone else has mentioned that so far.  I'm not sure I'm a fan, because it's one more thing to change when moving from a short to long form body.  Though in practice it should be rare, since 90% of the time you won't need to specify the argument anyway and can rely on the default.  (The only reason to specify it is to widen the type, frankly.)  Anyone else have an opinion here?

There's no real "proper" introduction.  Just github issues, and the Discord channel.  You're welcome to stop by there and chat if you are so inclined.  (Link is in the footer of the FIG website.)  I try to keep any formal decision making to issues, but obviously that's always imperfect.

spec.md
```diff
@@ -772,6 +772,171 @@ A function or method may be referenced in a way that creates a closure out of it
...
+```
+
+Property hooks also support multiple short-hook variations.
+
+For a `set` hook, if the argument name and type do not need to be redefined, then they MAY be omitted.
+
+If a hook consists of a single expression, then PHP allows it to be shortened using `=>`.  In that case:
+
+* There MUST be a single space on either side of the `=>` symbol.
+* The body MUST be on the same line as the hook declaration and MUST NOT be wrapped. If wrapping is necessary to keep the line within the line length or desired for other reasons the short-hook syntax MUST NOT be used.
```
Crell, line 831: Having spent more time working with hooks myself now, I am not sure I agree with this line anymore.  Ternaries are perfectly reasonable as a single-expression hook body, and those are often multilined for clarity.  I can certainly see match statements making sense in some cases, and those are obviously multi-line.  

I think I want to soften this to "must start on the same line", and then let other rules around how to wrap things apply.  Comments?

theodorejb: I agree. It's a similar situation to short closures, which also sometimes have an expression which it makes sense to split across multiple lines.

mathroc:
> Actually here's a better example from the same project:
> 
> ```
> class PageRecord
> {
>     public int $title {
>         get => $this->values(__PROPERTY__)[0];
>     }
> 
>     public int $order {
>         get => \max($this->values(__PROPERTY__));
>     }
> 
>     public bool $hidden {
>         get => array_all($this->values(__PROPERTY__), static fn($x): bool => (bool)$x);
>     }
> 
>     public bool $routable {
>         get => array_any($this->values(__PROPERTY__), static fn($x): bool => (bool)$x);
>     }
> 
>     public bool $isFolder {
>         get => array_any($this->values(__PROPERTY__), static fn($x): bool => (bool)$x);
>     }
> 
>     public \DateTimeImmutable $publishDate {
>         get => \max($this->values(__PROPERTY__));
>     }
> 
>     public \DateTimeImmutable $lastModifiedDate {
>         get => \max($this->values(__PROPERTY__));
>     }
> 
>     public string $pathName {
>         get => substr($this->logicalPath, strrpos($this->logicalPath, '/') + 1);
>     }
> 
>     public array $tags {
>         get => array_values(array_unique(array_merge(...$this->values(__PROPERTY__))));
>     }
> 
>     public function __construct(
>         public string $logicalPath,
>         public string $folder,
>         public array $files,
>     ) {}
> 
>     private function values(string $property): array
>     {
>         return array_column($this->files, $property);
>     }
> }
> ```
> 
> Adding another line break in there gives me no value, it just makes me scroll more.

I generally prefer having the curly brace on the declaration line (to me it feels like it's always just about scrolling more :wink:) but since the convention is otherwise to have on its own line this feels a bit weird

spec.md
Crell: Absent any feedback, I went ahead and made this change.

Crell:
Merged as there didn't seem to be any other feedback.

Issue #89, opened 2024-07-12
New PHP 8.4 Syntax: new MyClass()->method() without parentheses


KorvinSzanto:
[`new Foo()->baz()`](https://wiki.php.net/rfc/new_without_parentheses) will be allowed in PHP 8.4, we should cover it in the next version of PER-CS.

To me this is similar to the question of `new Foo` vs `new Foo()` where we decided the latter was better for readability: https://github.com/php-fig/per-coding-style/blob/master/spec.md#4-classes-properties-and-methods

Crell:
What should we say here?  "Favor using the non-parens version if you can", or is there more to say on the matter?

KorvinSzanto:
I'm of the opinion that it's simpler to see at a glance that a new instantiation is happening with wrapping parenthesis, but I don't think it's so much of an improvement that it justifies adding the wrapper in every situation.

Given that this is technically existing syntax, we'd need to go to `3.0` (or drop the thinking in #10) in order to add `MUST NOT use parenthesis` though we could require the opposite opposite since that's backwards compatible. In 2.1 we could add something like:

> When chaining new instantiations with method calls (IE `new Foo()->baz()`, the instantiation - including the new keyword - SHOULD NOT be wrapped in parenthesis. When dealing with a complex statement, wrapping parenthesis MAY be used for readability.

Crell:
As usual, I don't think "never use this fancy new syntax that was just approved" is going to fly.  The RFC passed 25 to 4. :-)  So let's take "MUST NOT" use it off the table, as it's clear people want to use it.

The question is whether we encourage it, or stay silent.  I'm leaning toward silent at the moment, but could be convinced otherwise.  (We can always say something about it in a future version.)

KorvinSzanto:
> As usual, I don't think "never use this fancy new syntax that was just approved" is going to fly. The RFC passed 25 to 4. :-) So let's take "MUST NOT" use it off the table, as it's clear people want to use it.

I agree, but I think you're misinterpreting my comment. I said "MUST NOT use parenthesis" which would mean we advocate for the thing that passed 25 to 4.

I'm saying that we cannot advocate stronger than a SHOULD until 3.0 unless we advocate for the thing that requires no code change, which I agree we shouldn't do.

Crell:
Oh!  I see what you mean.  I'm open to a SHOULD here if the rest of the WG is.

samdark:
@Crell sounds more weighted than MUST for the case :+1:

[...]

PR #110, opened 2025-01-21
Recommend omitting wrapping parenthesis when immediately calling a method on a new instance

KorvinSzanto:
The wording for this was proving tricky so I've included a counter example to make it more obvious what we're recommending be omitted.

Relates to #89

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -308,6 +308,19 @@ separated by a space. For example:
 class MyException extends \RuntimeException {}
 ```
 
+When calling a method immeditely after instantiating a new class, the instantiation SHOULD NOT be wrapped in 
+parenthesis. For example:
+
+```php
+$methodResult = new Foo()->someMethod();
+```
+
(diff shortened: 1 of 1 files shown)


Crell (approved):

Crell:
Looks good to me, but giving it a few days for the rest of the WG to approve/not.

samdark (approved):

spec.md
```diff
@@ -308,6 +308,23 @@ separated by a space. For example:
 class MyException extends \RuntimeException {}
 ```
 
+When accessing a class member immeditely after instantiating a new class, the instantiation SHOULD NOT be wrapped in
```
devfrey, line 311: There's a small typo here, should be "immediately".

Crell: Sigh.  Since it's already merged, can you send a tiny PR?  (I'm about to get on a plane.)

PR #111, opened 2025-01-21
Switch from RFC 2119 to BCP-14 which includes RFC 8174

KorvinSzanto:
[RFC 8174](https://datatracker.ietf.org/doc/html/rfc8174/) clears up ambiguity left in [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119/) by requiring any important keywords to be capitalized. This PR updates the wording at the top to match (mostly) what is specified by RFC 8174


Crell (approved):

Crell:
Thanks!

PR #113, opened 2025-05-13
Undo capitalization of `// no break` to prevent tooling issues

vdelau:
In [PR 79](https://github.com/php-fig/per-coding-style/pull/79) all comments where either replaced with ellipsis or capitalized. The `// no break` comment however is used literally in various tools and capitalizing this might lead to tooling regressions.

For reference, literal `// no break` seems to be a de-facto standard, even if it is not explicitly phrased as such: 
- https://youtrack.jetbrains.com/issue/WI-32123/Usage-of-PSR-2-comment-suggestion-on-no-break-switch-cases
- https://cs.symfony.com/doc/rules/control_structure/no_break_comment.html && https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/v3.75.0/src/Fixer/ControlStructure/NoBreakCommentFixer.php#L116
- https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#clever-code

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1092,7 +1092,7 @@ A `switch` structure looks like the following. Note the placement of
 parentheses, spaces, and braces. The `case` statement MUST be indented once
 from `switch`, and the `break` keyword (or other terminating keywords) MUST be
 indented at the same level as the `case` body. There MUST be a comment such as
-`// No break` when fall-through is intentional in a non-empty `case` body.
+`// no break` when fall-through is intentional in a non-empty `case` body.
 
 ```php
 <?php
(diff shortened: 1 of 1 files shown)


lpd-au (approved):

Crell (approved):

jrfnl:
For the record: the WordPress Coding Standards just look for _a comment_ without a predefined contents. Quoting that as a reference to justify this change is not correct and shows this hasn't been validated/tested.

lpd-au:
> For the record: the WordPress Coding Standards just look for _a comment_ without a predefined contents. Quoting that as a reference to justify this change is not correct and shows this hasn't been validated/tested.

The spec likewise only requires _a comment_, all this PR did was revert an unpublished change to the example (so it once again matches WordPress' example and works with PhpStorm + other tools).

PR #115, opened 2025-05-31
Clarify what visibility must be declared.

### LINKED ISSUE #114: Clarify "Visibility must be declared" in the context of aviz
Related to #98 and #99 

From the currently published PSR-PER (and not changed in `master`):

> 4.3 Properties and Constants[¶](https://www.php-fig.org/per/coding-style/#43-properties-and-constants)
> Visibility MUST be declared on all properties.

For the purpose of automated tooling and keeping aviz in mind, should the above be interpreted to mean "the _get/read_ visibility must always be explicitly declared, independently of a potentially declared _write_ visibility" ? Or should it be interpreted to mean "as long as there is _any_ visibility declared, be it aviz or non-aviz, this rule is satisfied" ?

Crell:
Resolves #114.

This is an obvious change so I'm going to go ahead and commit it.  The PR is just for historical reference and transparency.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -519,7 +519,7 @@ class Talker
 
 ### 4.3 Properties and Constants
 
-Visibility MUST be declared on all properties.
+Visibility MUST be declared on all properties.  If set-visibility is specified, then the general visibility MAY be omitted.
 
 Visibility MUST be declared on all constants.
 


kenguest (approved):
LGTM

ralflang (approved):

spec.md
```diff
@@ -519,7 +519,7 @@ class Talker
 
 ### 4.3 Properties and Constants
 
-Visibility MUST be declared on all properties.
+Visibility MUST be declared on all properties.  If set-visibility is specified, then the general visibility MAY be omitted.
```
KorvinSzanto, line 522:
```suggestion
Visibility MUST be declared on all properties. If set-visibility is specified, then the general visibility MAY be omitted.
```

Double spaces after a period aren't rendered by markdown unless they are followed by a newline. Most editors agree that two spaces after a period doesn't improve readability past the printed era.

PR #117, opened 2025-06-11
Trim out useless metadoc legacy

Crell:
This is mostly old PSR-2 justifications that have not been relevant for many years.  There's no reason for us to carry that baggage around.


GGOemea (commented):
Go

PR #118, opened 2025-06-16
Fix renumbering

Crell:
Missed a few earlier.


KorvinSzanto (approved):
LGTM though I'm not sure how the effects the readiness vote.

Crell:
It's non-substantive, so it doesn't.

PR #122, opened 2025-09-10
Spec section 4: static properties need a `$`

DanielEScherzer:
Fix example from `new Foo()::someStaticProperty;` to `new Foo()::$someStaticProperty;`


DanielEScherzer:
The static method call example is also problematic but technically not incorrect, so I sent a separate patch to fix that, #123

DanielEScherzer:
There is probably going to be a merge conflict between this and #123, whichever is merged first I'll rebase the other

samdark (approved):

PR #123, opened 2025-09-10
Spec section 4: static methods should be called statically

DanielEScherzer:
Technically they work fine when called non-statically, but doing so is generally discouraged

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -374,7 +374,7 @@ parentheses. For example:
 
 ```php
 new Foo()->someMethod();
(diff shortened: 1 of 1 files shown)


KorvinSzanto (approved):
Given that this method is being called on an instance, either variant works fine and is perfectly allowed. That said, we already demonstrate `->someMethod()` so we'd be better off using the paamneku.

samdark (approved):

Crell:
Thanks, @DanielEScherzer!

PR #130, opened 2025-11-19
The PER shouldn't refer to itself as a PSR

kenguest:
There are some occurrences in PER-CS where it refers to itself as "This PSR". These occurrences should be changed to "this standard"

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -224,7 +224,7 @@ $result = match ($a) {
 
 ### 2.7 Naming
 
-This STANDARD RECOMMENDS following the [php-src coding standards](https://github.com/php/php-src/blob/master/CODING_STANDARDS.md#user-functionsmethods-naming-conventions) with regard to abbreviations and acronyms.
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -224,7 +224,7 @@ $result = match ($a) {
 
 ### 2.7 Naming
 
-This PSR RECOMMENDS following the [php-src coding standards](https://github.com/php/php-src/blob/master/CODING_STANDARDS.md#user-functionsmethods-naming-conventions) with regard to abbreviations and acronyms.
+This STANDARD RECOMMENDS following the [php-src coding standards](https://github.com/php/php-src/blob/master/CODING_STANDARDS.md#user-functionsmethods-naming-conventions) with regard to abbreviations and acronyms.
```
Crell:
```suggestion
This Standard RECOMMENDS following the [php-src coding standards](https://github.com/php/php-src/blob/master/CODING_STANDARDS.md#user-functionsmethods-naming-conventions) with regard to abbreviations and acronyms.
```

No need to all-caps here. :smile:

kenguest: That's done now

Issue #135, opened 2026-02-07
Attributes on anonymous classes


KorvinSzanto:
Today our requirements for attributes mixed with the syntax required for anonymous classes gives us the rather confusing:

```php
$instance = new
#[Foo]
#[Baz]
#[Bar('foo', 'baz')]
class {};
```

https://3v4l.org/CM0Il

I propose we require an indent past the initial new to make it more obvious what's happening here:

```php
$instance = new
    #[Foo]
    #[Baz]
    #[Bar('foo', 'baz')]
    class {};
```
https://3v4l.org/8Hv0t

PR #136, opened 2026-02-07
Specify anonymous class attributes

KorvinSzanto:
See #135 

This PR adds to the anonymous class section information about how to handle attributes on anonymous classes.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1561,7 +1561,7 @@ $instance = new class ($a) extends \Foo implements
 
 If an anonymous class has attributes, they MUST be placed starting on the next line after the `new` keyword, indented
 once, and otherwise follow the same rules as other attributes specified in section 12. The remaining anonymous class 
-declaration after the final attribute MUST keep the same indentation as the attributes:
+declaration after the final attribute MUST start on a new line and keep the same indentation as the attributes:
 
 ```php
 $example = new
--- a/spec.md
+++ b/spec.md
@@ -1559,6 +1559,18 @@ $instance = new class ($a) extends \Foo implements
 };
 ```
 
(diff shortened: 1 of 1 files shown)


samdark (approved):

Crell:
Oh geez.  Attributes on anon classes are wild. :-)  Putting the attribute between the `new` and `class` feels gross, but it's a parse error otherwise, so, eh.

I'm OK with this.  Tagging @jrfnl for confirmation we're not creating any new problems.

spec.md
```diff
@@ -1559,6 +1559,18 @@ $instance = new class ($a) extends \Foo implements
 };
 ```
 
+If an anonymous class has attributes, they MUST be placed starting on the next line after the `new` keyword, indented
+once, and otherwise follow the same rules as other attributes specified in section 12. The remaining anonymous class 
+declaration after the final attribute MUST keep the same indentation as the attributes:
```
jrfnl:
```suggestion
declaration after the final attribute MUST start on a new line and keep the same indentation as the attributes:
```

jrfnl (commented):
@Crell Thanks for the ping. The wording seems clear, though I left one small suggestion.

KorvinSzanto:
> Oh geez. Attributes on anon classes are wild. :-) Putting the attribute between the `new` and `class` feels gross, but it's a parse error otherwise, so, eh.

Yeah I figure there really isn't an optimal place to put it other than maybe before the `$example = `, but that could be big trouble if PHP ever adds variable level attributes.

Crell:
It's also a syntax error at present. I tried. :smile:

PR #137, opened 2026-05-05
Fix #124 Require parenthesis around clone()

### LINKED ISSUE #124: PHP 8.5: Always require parentheses around `clone()`
Following #109.

With PHP 8.5 `clone()` is now a proper function that still supports a parentheses-less syntax for compatibility purposes, but requires parens for the new clone-with functionality. For consistency it should always be used with parentheses.

Correct:
```
$cloned = clone($obj);
$cloned = clone($obj, $with);
$cloned = clone($obj, ['prop' => $prop]);
$cloned = clone($obj, [
    'prop' => $prop,
]);
$cloned = \clone($obj);

function_call(clone($obj), $value);
```

Incorrect:
```
$cloned = clone $obj;
function_call(clone $obj, $value);
```

mbeccati:
Per request from @Crell on Discord ;-)

Resolves #124

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -853,6 +853,16 @@ die();
 $result = foo() ?? exit();
 ```
 
+The `clone()` function SHOULD always be called with parenthesis, even when the optional `$withProperties` argument is
+not provided. For example:
+
+```php
+$b = clone($a);
+$b = clone($a, [
+    'foo' => 'bar',
+]);
+```
+
 ### 4.8 Function Callable References
 
 A function or method may be referenced in a way that creates a closure out of it, by providing `...` in place of arguments.


spec.md
```diff
@@ -853,6 +853,16 @@ die();
 $result = foo() ?? exit();
 ```
 
+The `clone()` function SHOULD always be called with parenthesis, even when the optional `$withProperties` argument is
```
Crell: Why a SHOULD?  I would expect a MUST here, since it becomes a language-enforced MUST eventually.

Crell (approved):
I'm good with this.  Waiting a few days for other WG members to comment if desired.

spec.md
```diff
@@ -853,6 +853,16 @@ die();
 $result = foo() ?? exit();
 ```
 
+The `clone()` function MUST always be called with parenthesis, even when the optional `$withProperties` argument is
```
jrfnl:
```suggestion
The `clone` language construct MUST always be called with parenthesis, even when the optional `$withProperties` argument is
```

`clone` is not a function but a language construct, otherwise it couldn't be called without parenthesis....

mbeccati: Yes, my first draft used *language construct*, but then I checked the RFC and I found:

> We propose to change clone from a standalone keyword to a language construct that optionally accepts a second [array](http://www.php.net/array) parameter.

and:

> By promoting clone to a function, it will also be possible to use it as a callable. E.g., in [array_map](http://www.php.net/array_map). Promoting clone to a function also simplifies the implementation considerably.

Being consistent with the paragraph about `exit()` and `die()`, I chose to use *function* here too.

mbeccati:
@jrfnl I'm not a native speaker, but to me it feels *used* would be more appropriate than *called* when referring to a language construct (vs function). Thoughts?

jrfnl:
> @jrfnl I'm not a native speaker, but to me it feels _used_ would be more appropriate than _called_ when referring to a language construct (vs function). Thoughts?

@mbeccati I agree, `used` reads better.

jrfnl:
**Question for the interpretation by tooling:**

The examples have no spacing between the keyword and the open parenthesis and on the inside of the parentheses.

Can we conclude that the rule about this for function/method calls (line 781 - 785) also applies to these language constructs ?

> When making a method or function call, there MUST NOT be a space between the
method or function name and the opening parenthesis, there MUST NOT be a space
after the opening parenthesis, and there MUST NOT be a space before the
closing parenthesis. In the argument list, there MUST NOT be a space before
each comma, and there MUST be one space after each comma.

spec.md
```diff
@@ -853,6 +853,16 @@ die();
 $result = foo() ?? exit();
 ```
 
+The `clone` language construct MUST always be used with parenthesis, even when the optional `$withProperties` argument is
+not provided. For example:
```
mbeccati, line 857: @jrfnl How about something along the lines of

```suggestion
not provided. As such, it follows the same rules as function calls. For example:
```

jrfnl: I'm not sure it needs to be spelled out as this paragraph is within the "function and method call" section anyway, but I just wanted to clarify to avoid ending up with discussions about this in the PHPCS repo later on.

Crell:
Agreed with @jrfnl about the spacing.  Pseudo-function language constructs should always be formatted like functions, without a space.

mbeccati:
> print is not a function but a language construct. Its argument is the expression following the print keyword, and is not delimited by parentheses. 

and

> Note: Because this is a language construct and not a function, it cannot be called using [variable functions](https://www.php.net/manual/en/functions.variable-functions.php), or [named arguments](https://www.php.net/manual/en/functions.arguments.php#functions.named-arguments).

Yet, it has a return value and disguises as a function.

Should we also globally regulate that language constructs masking as functions should follow the same rules as functions?

Crell:
I'm tempted to say yes, but that would also mean always requiring `print` have parens, which I don't think anyone does currently.  So mandating that globally wouldn't be advisable.

jrfnl:
> I'm tempted to say yes, but that would also mean always requiring `print` have parens, which I don't think anyone does currently. So mandating that globally wouldn't be advisable.

Print is a "pure" language construct, so the parentheses have no function, other than to group whatever is being printed.
Exit/die/clone have over the last few years been changed to "language construct masking as functions" (which allows for fully qualified use, named arguments etc).

To me, there is a difference, though I can imagine not everyone can see that difference or is aware of it.

Crell:
Hm.  How do we define "language construct cosplaying a function" in the spec in a way that is clear to people?

mbeccati:
Function-like language constructs, such as...

Crell:
Will the casual PHP reader understand that `exit()` is a function-like language construct but `print()` is not?  Both permit parentheses.

jrfnl:
`print` allows parentheses in the same way as `return` or `include` allow parentheses. The parentheses do not have a function for the construct, they only group whatever comes after.

Maybe it should be something to be clarified/defined in section 2 ? And then other sections can refer to that definition ?

mbeccati:
With print, I'm a bit confused.

How do you suggest we treat the following?

```php
$i = print("foo");
```

Likewise:

```php
if (include("lib.php")) {
    ...
}
```

KorvinSzanto:
This feels a bit like scope creep to me. Can't we just add clone to the SHOULD list directly above that includes `exit` and `die` and work on a more general rule for language constructs in a separate issue? I'd argue we could get by with clone parenthesis being a SHOULD for now.

Crell:
That... does seem like a reasonable and low-impact approach. :smile: Any objections?

mbeccati:
Well, the explanation given in the paragraph doesn't really apply to `clone`, which can never be mistaken for a constant.

The following makes little sense to me:

> The `exit()`, `die()`, and `clone()` functions SHOULD always be called with parentheses even if no argument is given to clearly distinguish them from an access to a constant named `exit` or `die`.

TimWolla:
> Print is a "pure" language construct, so the parentheses have no function, other than to group whatever is being printed.

This is not quite correct: `print` is an expression with super low precedence.

So:

```
<?php

var_dump(print(1) + 2);
```

is actually printing `3` and then var_dump'ing `int(1)`, because that's what the result of the `print` expression is.

`clone()` is special in that it already had the highest precedence of all expressions so that it only captures the shortest expression that follows, which means that adding parentheses is not changing anything in the parse tree.

And with 8.5 it actually is a full proper function, just a function that has a little bit of a special treatment in the parser, but nowadays saying “clone is a function” would be closer to being correct than saying “clone is a language construct”. And the same is true for `exit()` and `die()` (which always required parens when passing a parameter).

mbeccati:
Following the feedback, I've reverted back to my original version with `function` and `SHOULD`.

Crell (approved):
Good enough at this point.  Thanks.

Issue #138, opened 2026-05-15
Section links in migration-2.0.md point to PER-CS 3.0 instead of 2.0


rodrigoprimo:
While working on an upcoming PR suggesting a change to `migration-2.0.md`, I noticed that the section headings link to the unversioned URL for the spec (https://www.php-fig.org/per/coding-style/). This URL currently serves PER-CS 3.0. I believe they should point to sections in the PER-CS 2.0 spec file (https://github.com/php-fig/per-coding-style/blob/2.0.0/spec.md). Is this correct or am I missing something?

Happy to create a PR to update the links if this is indeed an issue.

Crell:
I think there's a valid question here: Do we want to link to "what you should be doing" (latest version) to review in total, or to the possibly-outdated "here's the one change?"

Thinking forward into the future where we have 6 different older versions, linking to the specific version could lead someone to try to adopt a now-outdated rule.  OTOH, linking to the current version could get really complicated if we ever make more substantial changes to the spec, like reorganizing things or more BC-breaking changes.

I'm not sure here.  Open to thoughts from the WG.

rodrigoprimo:
Thanks for your input. Both concerns are fair, and I'm happy to defer to what you and the rest of the WG prefer.

> Thinking forward into the future where we have 6 different older versions, linking to the specific version could lead someone to try to adopt a now-outdated rule.

This is something I hadn't considered. To be honest, I don't see this as a big concern, since I think the migration document itself mitigates the risk. Specifically, `migration-2.0.md` is scoped to the 1.0 to 2.0 transition, so a reader using it is explicitly looking at that upgrade.

> OTOH, linking to the current version could get really complicated if we ever make more substantial changes to the spec, like reorganizing things or more BC-breaking changes.

This is the main concern I had in mind when I opened this issue.

Crell:
Pinging the rest of the working group to weigh in here...  This blocks a half-dozen PRs to improve the migration guides.

@rodrigoprimo Would you also be interested in taking a crack at the 3.1 migration guide, since it looks like we're done with any changes there?  Since you seem to have a good eye for changes. 😄

KorvinSzanto:
I do think we should probably link to contemporary copies in the migration guide. We probably also should be working on a 3.0 migration guide rather than modifying the existing 1-2 guide.

Crell:
Which are contemporary? 😄 

1. The current version.
2. The version that is the upgrade target of the particular migration guide.

[...]

PR #139, opened 2026-05-15
Add Section 6.1 to the PER-CS 1.0 to 2.0 migration guide

rodrigoprimo:
PER-CS 2.0 introduces a new normative rule in [Section 6.1](https://github.com/php-fig/per-coding-style/blob/2.0.0/spec.md#61-unary-operators):

> Type casting operators MUST NOT have any space within the parentheses and MUST be separated from the variable they are operating on by exactly one space.

The clause "and MUST be separated from the variable they are operating on by exactly one space" was added by #40 and is not present in [Section 6.1 in 1.0](https://github.com/php-fig/per-coding-style/blob/1.0.0/spec.md#61-unary-operators). I'm new to this repository, so I might be missing something, but this change is not documented in the 2.0 migration guide, and I believe it should be.

This PR adds a Section 6.1 entry to `migration-2.0.md`.

The new Section 6.1 heading uses the same unversioned `php-fig.org` URL as the existing entries. This URL currently points to PER-CS 3.0. I'm not sure if this is correct. I opened https://github.com/php-fig/per-coding-style/issues/138 to discuss it.

### DIFF
--- a/migration-2.0.md
+++ b/migration-2.0.md
@@ -138,6 +138,15 @@ $returnValue = match ($expr) {
 };
 ```
 
+## [Section 6.1 - Unary operators](https://www.php-fig.org/per/coding-style/#61-unary-operators)
+
+The type casting operators rule was updated to also require a single space between the cast and the variable it operates on.
+
+```php
+<?php
+$intValue = (int) $input;
+```
+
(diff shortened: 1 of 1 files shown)


Crell (approved):
This seems fine to me.  Will leave open for a bit to see if anyone else from the WG disagrees.

PR #140, opened 2026-05-15
lapsus: s/intersection/union/ with null

claudepache:
An intersection of a type with `null` is semantically either `null` or `never` (depending on whether the type is nullable). But it is not allowed by PHP anyway.


Jean85 (approved):
Thank you for spotting this!

Crell (approved):

PR #142, opened 2026-05-27
Add Section 8 to the PER-CS 1.0 to 2.0 migration guide

rodrigoprimo:
PER-CS 2.0 introduces a new rule in [Section 8](https://github.com/php-fig/per-coding-style/blob/2.0.0/spec.md#8-anonymous-classes):

> If the anonymous class has no arguments, the `()` after `class` MUST be omitted.

This rule was added by #37 and is not present in [Section 8 in 1.0](https://github.com/php-fig/per-coding-style/blob/1.0.0/spec.md#8-anonymous-classes).

This PR adds a Section 8 entry to `migration-2.0.md`.

As with #139, the new Section 8 heading uses the unversioned `php-fig.org` URL to match the existing entries. See #138 for the related discussion about whether this URL should be versioned.

### DIFF
--- a/migration-2.0.md
+++ b/migration-2.0.md
@@ -167,6 +167,19 @@ $func = fn(
 $result = $collection->reduce(fn(int $x, int $y): int => $x + $y, 0);
 ```
 
+## [Section 8 - Anonymous Classes](https://github.com/php-fig/per-coding-style/blob/2.0.0/spec.md#8-anonymous-classes)
+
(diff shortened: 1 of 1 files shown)


samdark (approved):

rodrigoprimo:
Rebased onto the latest master and force-pushed. This picks up #152, which (following the decision in #138) switched the migration guide links to point to the target version on GitHub instead of the website. I also updated the link this PR adds to follow the same convention.

Crell (approved):

PR #64, opened 2023-03-01
Document considered but unresolved scope clarifications for 2.0 in the meta document

lpd-au:
Supercedes https://github.com/php-fig/per-coding-style/pull/63 for 2.0.0.

This pull request:

- Clarifies that this PER obsoletes PSR-12 but extends PSR-1.
- Documents our considered but unresolved discussions around scope and PSR-1.

PR #152, opened 2026-06-22
Point migration guide links to the target version on GitHub

rodrigoprimo:
Following the decision in #138, all links in the migration guides should point to the target version on GitHub rather than the current version on the website, which always reflects the latest spec and can drift away from what the guide describes.

This PR updates migration-2.0.md to link to spec.md at the 2.0.0 tag and migration-3.0.md to link to spec.md at the 3.0.0 tag, covering both the per-section links and the canonical-source link in each intro.

Note that the Section 6.4 anchor in migration-3.0.md changes its fragment from `#64-operator-placement` to `#64-operators-placement` to match the heading on GitHub.

### DIFF
--- a/migration-2.0.md
+++ b/migration-2.0.md
@@ -17,10 +17,10 @@ and focuses on the changes on a section-by-section basis as its focus is to be m
 This document intends to provide a summary of these changes that can
 then be used to drive action lists for toolset producers to support PER-CS v2.0.
 
-This document is non-normative.  The published [2.0 PER-CS](https://www.php-fig.org/per/coding-style/) specification 
+This document is non-normative.  The published [2.0 PER-CS](https://github.com/php-fig/per-coding-style/blob/2.0.0/spec.md) specification 
 is the canonical source for the PER-CS formatting expectations.
 
-## [Section 2.6 - Trailing Commas](https://www.php-fig.org/per/coding-style/#26-trailing-commas)
+## [Section 2.6 - Trailing Commas](https://github.com/php-fig/per-coding-style/blob/2.0.0/spec.md#26-trailing-commas)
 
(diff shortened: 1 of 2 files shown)


Crell:
I probably caused the conflict here with the merge I literally just did, sorry. :grimacing: Can you rebase?

migration-2.0.md
```diff
@@ -17,10 +17,10 @@ and focuses on the changes on a section-by-section basis as its focus is to be m
 This document intends to provide a summary of these changes that can
 then be used to drive action lists for toolset producers to support PER-CS v2.0.
 
-This document is non-normative.  The published [2.0 PER-CS](https://www.php-fig.org/per/coding-style/) specification 
+This document is non-normative.  The published [2.0 PER-CS](https://github.com/php-fig/per-coding-style/blob/2.0.0/spec.md) specification
```
Crell, line 20: Hm, should this be an exception that still points to the latest release?

rodrigoprimo:
> I probably caused the conflict here with the merge I literally just did, sorry. 😬 Can you rebase?

@Crell, it wasn't you. I forgot to update the `master` on my fork before creating this branch, so it did not include the changes from #139 that were merged a while ago. I just rebased this branch with the latest from `master` and force-pushed.

migration-2.0.md
rodrigoprimo: I don't think so. The reason is that the link name mentions the specific PER-CS version, which is why I also opted to point to the specific version on GitHub instead of the latest release at https://www.php-fig.org/per/coding-style/. Unless we want to change the link name as well to make it clearer that it is a link to the latest PER-CS version?

Crell: Valid.  OK, we'll roll with it.

Crell (approved):

PR #155, opened 2026-08-12
Fix inconsistencies and invalid examples in specification

samdark:
This fixes several inconsistencies and invalid examples in the specification.

Changes include:

- use an explicit nullable parameter and correct the backed-enum example
- clarify parentheses for intersections within union types
- clarify import-name resolution
- make the `declare` spacing rule apply to all directives while presenting `strict_types` as an example
- correct indentation in code examples
- use the normative BCP 14 keyword for the switch requirement
- make the multiline-array rule applicable outside assignments
- document the expanded multiline-array rule in the v3.1 migration guide

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -56,7 +56,7 @@ use const Vendor\Package\{ConstantA, ConstantB, ConstantC};
 
 class Foo extends Bar implements FooInterface
 {
-    public function sampleFunction(int $a, int $b = null): array
+    public function sampleFunction(int $a, ?int $b = null): array
     {
         if ($a === $b) {
             bar();
@@ -80,7 +80,7 @@ enum Beep: int
 
     public function isOdd(): bool
     {
-        return $this->value() % 2;
+        return $this->value % 2 !== 0;
     }
 }
 ```
@@ -132,9 +132,9 @@ Any new types and keywords added to future PHP versions MUST be in lower case.
 Short form of type keywords MUST be used i.e. `bool` instead of `boolean`,
 `int` instead of `integer` etc.
 
-Compound types includes intersection, union, and mixed intersection and union type declarations. PHP requires
-that all compound types be structured as an ORed (unioned) series of ANDs (intersections), and that each set of
-intersections be encased with parentheses.
+Compound types include intersection, union, and mixed intersection and union type declarations. PHP requires
+that mixed intersection and union types be structured as an ORed (unioned) series of ANDs (intersections),
+and that each intersection in such a union be encased with parentheses.
 
 The union symbol `|` and intersection symbol `&` MUST NOT have a leading or trailing space.  The parentheses MUST NOT
 have a leading or trailing space.
@@ -160,7 +160,7 @@ function somethingWithReflection(
     |\ReflectionParameter
     |\ReflectionProperty $reflect
 ): object|null {
-        // ...
+    // ...
 }
 
 function complex(array|(ArrayAccess&Traversable) $input): ArrayAccess&Traversable
@@ -255,8 +255,8 @@ When the opening `<?php` tag is on the first line of the file, it MUST be on its
 own line with no other statements unless it is a file containing markup outside of PHP
 opening and closing tags.  The `<?php` tag MUST always be lower case.
 
-Import statements MUST never begin with a leading backslash as they
-must always be fully qualified.
+Import statements MUST never begin with a leading backslash, as imported names
+are always resolved from the namespace root.
 
 The following example illustrates a complete list of all blocks:
 
@@ -335,8 +335,8 @@ For example:
 </html>
 ```
 
-Declare statements MUST NOT contain any spaces and MUST be exactly `declare(strict_types=1)`
-(with an optional semicolon terminator).
+Declare statements MUST NOT contain any spaces inside the parentheses. For example:
+`declare(strict_types=1)` (with an optional semicolon terminator).
 
 Block declare statements are allowed and MUST be formatted as below. Note position of
 braces and spacing:
@@ -606,8 +606,8 @@ class Point
 class Point
 {
     public function __construct(
-      public readonly int $x,
-      public readonly int $y,
+        public readonly int $x,
+        public readonly int $y,
     ) {}
 }
 ```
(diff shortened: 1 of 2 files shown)


spec.md
```diff
@@ -56,7 +56,7 @@ use const Vendor\Package\{ConstantA, ConstantB, ConstantC};
 
 class Foo extends Bar implements FooInterface
 {
-    public function sampleFunction(int $a, int $b = null): array
+    public function sampleFunction(int $a, ?int $b = null): array
```
samdark, line 59: This syntax still works in PHP 8.4, but emits a deprecation notice. Using `?int` avoids the notice and follows the specification’s later recommendation for explicit nullable types.

spec.md
```diff
@@ -80,7 +80,7 @@ enum Beep: int
 
     public function isOdd(): bool
     {
-        return $this->value() % 2;
+        return $this->value % 2 !== 0;
```
samdark, line 83: Backed enum values are exposed through the `value` property, not a `value()` method. The explicit comparison also makes the expression produce the declared `bool` under `strict_types=1`.

spec.md
```diff
@@ -132,9 +132,9 @@ Any new types and keywords added to future PHP versions MUST be in lower case.
 Short form of type keywords MUST be used i.e. `bool` instead of `boolean`,
 `int` instead of `integer` etc.
 
-Compound types includes intersection, union, and mixed intersection and union type declarations. PHP requires
-that all compound types be structured as an ORed (unioned) series of ANDs (intersections), and that each set of
-intersections be encased with parentheses.
+Compound types include intersection, union, and mixed intersection and union type declarations. PHP requires
+that mixed intersection and union types be structured as an ORed (unioned) series of ANDs (intersections),
+and that each intersection in such a union be encased with parentheses.
```
samdark, line 137: A standalone intersection is written `A&B`, without parentheses. Parentheses are required only when that intersection is one option in a union, as in `A|(B&C)`. The previous wording incorrectly implied that standalone intersections should also be parenthesized.

spec.md
```diff
@@ -160,7 +160,7 @@ function somethingWithReflection(
     |\ReflectionParameter
     |\ReflectionProperty $reflect
 ): object|null {
-        // ...
+    // ...
```
samdark, line 163: This example is presented as correctly formatted, so its body should use the mandatory four-space indentation defined in section 2.4.

spec.md
```diff
@@ -255,8 +255,8 @@ When the opening `<?php` tag is on the first line of the file, it MUST be on its
 own line with no other statements unless it is a file containing markup outside of PHP
 opening and closing tags.  The `<?php` tag MUST always be lower case.
 
-Import statements MUST never begin with a leading backslash as they
-must always be fully qualified.
+Import statements MUST never begin with a leading backslash, as imported names
+are always resolved from the namespace root.
```
samdark, line 259: An imported name is resolved from the namespace root by PHP; calling it “fully qualified” is potentially confusing because fully qualified names are normally written with a leading backslash.

spec.md
```diff
@@ -606,8 +606,8 @@ class Point
 class Point
 {
     public function __construct(
-      public readonly int $x,
-      public readonly int $y,
+        public readonly int $x,
```
samdark, line 609: Constructor-promoted parameters are still parameters and therefore use the standard four-space indentation level.

spec.md
```diff
@@ -816,7 +816,7 @@ $foo->bar(
 <?php
 
 somefunction($foo, $bar, [
-  // ...
+    // ...
```
samdark, line 819: The array is a single multiline argument rather than a split argument list, but its contents still need the standard four-space indentation.

spec.md
```diff
@@ -1098,7 +1098,7 @@ if (
 
 ### 5.2 `switch`, `case`, `match`
 
-A switch structure must follow the rules below:
+A switch structure MUST follow the rules below:
```
samdark, line 1101: The BCP 14 preamble gives normative meaning only to uppercase keywords. Since the detailed switch rules are requirements, this introductory `MUST` should be normative too.

spec.md
```diff
@@ -1466,15 +1466,15 @@ $longArgs_noVars = function (
     $longerArgument,
     $muchLongerArgument,
 ) {
-   // ...
+    // ...
```
samdark, line 1469: These closure-body comments used three spaces while section 2.4 requires four. The same correction is applied to every example in this group.

spec.md
```diff
@@ -335,8 +335,8 @@ For example:
 </html>
 ```
 
-Declare statements MUST NOT contain any spaces and MUST be exactly `declare(strict_types=1)`
-(with an optional semicolon terminator).
+Declare statements MUST NOT contain any spaces inside the parentheses. For example:
+`declare(strict_types=1)` (with an optional semicolon terminator).
```
samdark, line 339: The no-spaces rule applies to every `declare` statement. The previous wording made `declare(strict_types=1)` sound like the only permitted declaration, even though it is just an example of the required formatting; other directives such as `declare(ticks=1)` are also allowed.

spec.md
```diff
@@ -1731,8 +1731,8 @@ Array declarations MAY be split across multiple lines, where each subsequent lin
 is indented once. When doing so, the first value in the array MUST be on the 
 next line, and there MUST be only one value per line.
 
-When the array declaration is split across multiple lines, the opening bracket 
-MUST be placed on the same line as the equals sign. The closing bracket 
+When an array declaration is split across multiple lines, the opening bracket
+MUST NOT be placed on its own line. The closing bracket
```
samdark, line 1735: The old wording mentioned an equals sign, but multiline arrays can also appear in arguments, return statements, defaults, attributes, and nested expressions. Requiring the opening bracket not to be on its own line makes the rule applicable in every context.

spec.md
Crell: Technically the previous text is correct: A single set of ANDs is still technically ORed with the empty set.  However, you're correct that the parens are only meaningful in mixed cases, so we'll go with these revisions.

Crell (approved):

